PR #12, opened 2022-06-13
AS-1061 fix slow jest

maxi-askui:
Added jest config for faster execution to example project jest.config.ts and documented it. Had to increase the major version of jest to be able to add the newest name of the config option to the example project (the name was changed between jest v27 and v28). Also added docu about the old option name.


adi-wan-askui (approved):
Awesome job! :)

PR #16, opened 2022-06-15
As 1007 troubleshooting UI controller wayland problems

thinh-askui:


packages/askui-nodejs/src/lib/control-ui-server-linux.ts
```diff
@@ -1,10 +1,32 @@
...
+    }
+    const checkLinuxdistrib = await runCommand('uname -v');
+
+    if (checkLinuxdistrib.stdout.trim().includes('Ubuntu')) {
+      try {
+        await runCommand('dpkg -s libfuse2 | grep Status');
+      } catch {
+        throw new LibfuseError('Libfuse2 package is missing: https://docs.askui.com/docs/general/Troubleshooting/askui-ui-controller-starting-problems#libfuse2');
+      }
+    }
```
maxi-askui, line 30: Could check for dpkg instead of "Ubuntu" so that other debian distributions e.g. Mint are also covered

packages/askui-nodejs/src/lib/ui-controller-linux.ts
```diff
@@ -1,10 +1,35 @@
...
+
+  // eslint-disable-next-line class-methods-use-this
+  protected override async preStartChecks(): Promise<void> {
+    const runCommand = promisify(exec);
+    const waylandStatus = await runCommand('echo $WAYLAND_DISPLAY');
+
+    if (waylandStatus.stdout.trim().includes('wayland')) {
+      throw new WaylandError('Wayland is not supported: https://docs.askui.com/docs/general/Troubleshooting/askui-ui-controller-starting-problems#wayland');
+    }
+
```
programminx-askui, line 22: Add comment why we are doing the check for dpkg and libfuse2

docs/docs/general/07-Troubleshooting/askui-ui-controller-starting-problems.md
```diff
@@ -0,0 +1,27 @@
+---
+custom_edit_url: null
+---
+
+# askui UI Controller Starting Problems
+
+## Wayland
+
+At the moment we don't support the windowing system Wayland for any linux distribution. 
+As solution you have to use Xorg instead of Wayland. You can look on [this page](https://www.maketecheasier.com/switch-xorg-wayland-ubuntu1710/) to see how to Switch to Xorg from Wayland.
```
mlikasam-askui, line 10:
```suggestion
As a solution, you can switch to Xorg instead of Wayland. [ how to switch to Xorg](https://www.maketecheasier.com/switch-xorg-wayland-ubuntu1710/) to see how to Switch to Xorg from Wayland.
```

docs/docs/general/07-Troubleshooting/askui-ui-controller-starting-problems.md
```diff
@@ -0,0 +1,27 @@
...
+
+
+
+## libfuse2 
+
+With Ubuntu 22.04 libfuse will no longer be installed per default.
+That's why you have to manually install libfuse2 to use the askui lib
+
+Therefore just run the following commands:
```
mlikasam-askui, line 19:
```suggestion
If you are using Ubuntu 22.04 or later, you need to install libfuse before using askui lib since it's no longer installed per default. 

libfuse can be installed with these commands: 
```

packages/askui-nodejs/README.md
```diff
@@ -42,6 +42,11 @@ npx askui init
 
 Visit our [documentation](https://docs.askui.com) for examples and a full list of supported commands.
 
+### Notes
+
+Important note for Linux users: Currently, Wayland is not supported.
+You can read more in our [troubleshooting chapter](https://docs.askui.com/docs/general/Troubleshooting/askui-ui-controller-starting-problems#wayland).
```
mlikasam-askui, line 48: well done, it will be nice also to include a hint to the libfuse problem

packages/askui-nodejs/src/lib/ui-controller-facade.ts
```diff
@@ -74,6 +75,18 @@ export abstract class UiControllerFacade {
...
+  protected preStartChecks(): Promise<void> {
+    return new Promise((resolve, reject) => {
+      try {
+        resolve();
+        return;
+      } catch (err) {
+        reject(new UnkownError(`An unknown error occured while waiting for the askui server: ${err}`));
+      }
+    });
+  }
```
mlikasam-askui, line 88: do you need this try and catch?

packages/askui-nodejs/src/lib/ui-controller-linux.ts
```diff
@@ -1,10 +1,35 @@
...
+    try {
+      await runCommand('dpkg --version');
+    } catch (err) {
+      return;
+    }
```
mlikasam-askui, line 27: can you please explain to me what is this doing?

thinh-askui: this is checking if we dpkg is used (if user has debian distribution). First i only checked if ubuntu is used. But Maxi and Dominik mentioned this problem can also occur on mint etc.

mlikasam-askui: sorry, my question was regarding the return;

thinh-askui: if it is not using dpkg/ no debian distribution. I want to end the check and don't want the libfuse2 control anymore
because `dpkg -s libfuse2 | grep Status` would always fail

packages/askui-nodejs/src/lib/ui-controller-facade.ts
```diff
@@ -74,6 +75,13 @@ export abstract class UiControllerFacade {
     /* Executable out of the box */
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  protected preStartChecks(): Promise<void> {
+    return new Promise((resolve) => {
+      resolve();
+    });
+  }
```
programminx-askui, line 83: What is this for a trick? 

The promise resolves directly. We don't need it.

programminx-askui: Ahh I see.

packages/askui-nodejs/src/lib/ui-controller-linux.ts
mlikasam-askui: then i assume you can remove this?

thinh-askui: I think I need this? method return if there is no dpkg and should not check for libfuse2

mlikasam-askui: so the final result we want to check for libfuse2. 
if `dpkg --version` fails, we can't check for libfuse2, and you assume that it will be there, and that's why you are using the return ?

docs/docs/general/07-Troubleshooting/askui-ui-controller-starting-problems.md
```diff
@@ -0,0 +1,28 @@
...
+At the moment we don't support the windowing system Wayland for any linux distribution. 
+As a solution, you can switch to Xorg instead of Wayland ([how to switch to Xorg](https://www.maketecheasier.com/switch-xorg-wayland-ubuntu1710/)). 
+
+
+
+## libfuse2 
+
+If you are using Ubuntu 22.04 or later, you need to install libfuse before using askui lib 
+since it's no longer installed per default.
+That's why you have to manually install libfuse2 to use the askui lib
```
mlikasam-askui, line 18: rm line 18, duplicated

docs/versioned_docs/version-0.1.6/general/07-Troubleshooting/askuicontroller-starting-problems.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+```shell
+sudo apt-get install libfuse2
+```
+
+
+
+
+
```
mlikasam-askui, line 32: file can be removed based on the discussion that we had in development channel

docs/docs/general/07-Troubleshooting/askui-ui-controller-starting-problems.md
```diff
@@ -0,0 +1,27 @@
...
+
+Libfuse2 can be installed with these commands: 
+
+```shell
+sudo apt-get update
+```
+
+```shell
+sudo apt-get install libfuse2
+
```
adi-wan-askui, line 27: Code block needs to be ended

packages/askui-nodejs/src/lib/libfuse-error.ts
```diff
@@ -0,0 +1,8 @@
+export class LibfuseError extends Error {
+  contructorName: string = this.constructor.name;
```
adi-wan-askui, line 2: Why? I have never seen this.

packages/askui-nodejs/src/lib/ui-controller-facade.ts
```diff
@@ -74,6 +75,13 @@ export abstract class UiControllerFacade {
     /* Executable out of the box */
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  protected preStartChecks(): Promise<void> {
+    return new Promise((resolve) => {
```
adi-wan-askui, line 80: Can we simplify this to `return Promise.resolve();`

packages/askui-nodejs/src/lib/ui-controller-linux.ts
```diff
@@ -1,10 +1,40 @@
...
+import { LibfuseError } from './libfuse-error';
 
 export class UiControllerLinux extends UiControllerFacade {
   // eslint-disable-next-line class-methods-use-this
   protected override makeBinaryExecutable(): void {
     exec(`chmod +x "${this.binaryPath}"`, (_exception, stdout) => logger.debug(stdout));
   }
+
+  // eslint-disable-next-line class-methods-use-this
+  protected override async preStartChecks(): Promise<void> {
```
adi-wan-askui, line 15: Don't like the name as it seems like something static but it is a function. What do you think about `runPreStartChecks()`?

packages/askui-nodejs/src/lib/ui-controller-facade.ts
adi-wan-askui: Like for the use of the pattern :)

adi-wan-askui (changes requested):
Nicely done :)

adi-wan-askui (approved):

PR #20, opened 2022-06-21
As 1073 improve speed if image resizing in lib

thinh-askui:


examples/askui-nodejs-base-example/package.json
```diff
@@ -7,17 +7,18 @@
         "lint:staged": "lint-staged"
     },
     "dependencies": {
-        "askui": "0.1.7"
+        "askui": "0.1.7",
```
adi-wan-askui, line 10: We don't have a version `0.1.7`, yet.

examples/askui-nodejs-base-example/package.json
```diff
@@ -7,17 +7,18 @@
...
+        "node-machine-id": "^1.1.12"
```
adi-wan-askui, line 11: I think this can and should be removed as this is installed as a transitive dependency.

packages/askui-nodejs/package.json
```diff
@@ -31,17 +31,20 @@
         "url-join": "~4.0.1",
         "wait-port": "~0.2.9",
         "ws": "~7.4.4",
-        "yup": "~0.32.11"
+        "yup": "~0.32.11",
+        "sharp": "^0.30.6"
```
adi-wan-askui, line 35: I think we can remove `jimp`.

packages/askui-nodejs/package.json
```diff
@@ -31,17 +31,20 @@
...
         "@types/fs-extra": "~9.0.13",
         "@types/jest": "~27.4.0",
         "@types/jsdom": "~16.2.14",
         "@types/node": "~17.0.10",
+        "@types/sharp": "^0.30.4",
         "@types/url-join": "~4.0.1",
         "@types/webrtc": "~0.0.30",
         "@types/ws": "~7.4.4",
         "jest": "28.1.1",
+        "sharp": "^0.30.6",
```
adi-wan-askui, line 47: Should only be included as a dependency.

packages/askui-nodejs/src/utils/transformations.ts
```diff
@@ -40,21 +40,26 @@ export async function resizeBase64ImageWithSameRatio(
   }
   try {
     let resizeRatio = 1;
-    const orignalImage = await read(Buffer.from(base64ImageString.replace(/^data:image\/png;base64,/, ''), 'base64'));
-    const imageHeight = orignalImage.getHeight();
-    const imageWidth = orignalImage.getWidth();
+    const orignalImage = sharp(Buffer.from(base64ImageString.replace(/^data:image\/png;base64,/, ''), 'base64'));
```
adi-wan-askui, line 43: Can we encapsulate the adding (line 12, 38, 62 etc.) and removing of the prefix, e.g., inside class?

packages/askui-nodejs/src/utils/transformations.ts
```diff
@@ -40,21 +40,26 @@ export async function resizeBase64ImageWithSameRatio(
...
+    const imageMetadata = await orignalImage.metadata();
+    const imageHeight = imageMetadata.height;
+    const imageWidth = imageMetadata.width;
 
-    if (Math.max(imageHeight, imageWidth) <= maxEdge) {
+    if (Math.max(imageHeight as number, imageWidth as number) <= maxEdge) {
```
adi-wan-askui, line 48: Do we really need the typecasts? If yes, they should be done as early as possible, i.e., on assignment of variables. This also allow multiple typecasts, e.g., in line 52 or 61.

packages/askui-nodejs/src/utils/transformations.ts
```diff
@@ -40,21 +40,26 @@ export async function resizeBase64ImageWithSameRatio(
...
+    const imageRatio = (imageHeight as number) / (imageWidth as number);
     const isHeigthBiggerThanWidth = imageRatio > 1;
     const newHeigh = isHeigthBiggerThanWidth ? maxEdge : maxEdge * imageRatio;
     const newWidth = isHeigthBiggerThanWidth ? maxEdge / imageRatio : maxEdge;
-    const newImage = orignalImage.resize(newWidth, newHeigh);
-    resizeRatio = imageHeight / newHeigh;
-    const newImageBase64 = await newImage.getBase64Async('image/png');
+
+    const newHeighint = Number.parseInt(newHeigh.toString(), 10);
+    const newWidthint = Number.parseInt(newWidth.toString(), 10);
```
adi-wan-askui, line 58: Can we make this numerically safer?

packages/askui-nodejs/src/utils/transformations.ts
```diff
@@ -8,8 +8,8 @@ export async function toBase64Image(imagePath: string): Promise<string> {
   if (!(fs.existsSync(imagePath))) {
     throw new Error(`the image ${imagePath} does not exists!`);
   }
-  const image = await read(imagePath);
-  return image.getBase64Async('image/png');
+  const image = sharp(imagePath);
+  return `data:image/png;base64,${(await image.toBuffer()).toString('base64')}`;
 }
 
 export async function toBase64ImageIfNeeded(pngPathOrBase64Image: string): Promise<string> {
```
adi-wan-askui, line 15: I think we can and should extract some functions as this is quite a long function including multiple steps.

adi-wan-askui (changes requested):
All in all, nice job introducing this. Should improve performance :) 

Some typos I found: 

- `newHeighint` ⟶ `newHeightInt`
- `orignalImage`⟶ `originalImage`
- `newHeigh` ⟶ `newHeight`
- `isHeigthBiggerThanWidth` ⟶ `isHeightBiggerThanWidth`

examples/askui-nodejs-base-example/package.json
mlikasam-askui: the reason why I upgraded the version, was that I already updated the example with the current changes, and the current version doesn't support that version

adi-wan-askui: I see but I don't like that it is pointing to version that doesn't exist. No one can actually run that example right now. What about pointing to `../../packages/askui-nodejs` for the time being.

adi-wan-askui: https://askyourui.atlassian.net/browse/AS-1081

mlikasam-askui: I would prefer then to rename it to the e2e test and point it to the `../../packages/askui-nodejs`. 
because it can't be made sure that it's up to date without first releasing a new version.
maybe also exclude updating it from the tickets. then we can keep the old version

adi-wan-askui: I agree with the last point. I would also go for keeping the old version and only upgrading examples after a new release. I think this would save us a lot of trouble.

Because of that I would see `../../packages/askui` as a temporary fix that we are going to replace with a proper version after the next release today. So I don't think renaming is necessary.

mlikasam-askui: the main drawback, if I want an example for a specific version won't be included inside the code when I check out the same tag number as the version

adi-wan-askui: Let's discuss it later after the Daily.

adi-wan-askui (approved):

PR #21, opened 2022-06-22
AS-920 Type Secrets

adi-wan-askui:
Adds methods for typing secrets that should neither be sent to inference backend nor be included in logs.


mlikasam-askui (approved):

packages/askui-nodejs/src/execution/execution-runtime.ts
```diff
@@ -84,18 +94,25 @@ export class ExecutionRuntime {
     return command;
   }
 
+  private async getImageIfRequired(instruction: string): Promise<string | undefined> {
+    const isImageRequired = await this.inferenceClient.isImageRequired(instruction);
+    if (!isImageRequired) return undefined;
```
programminx-askui, line 99: Does not return a promise of undefined.  --> Suprise

if condtion should contain always brackets for the body.

adi-wan-askui: My private coding style came through where I leave out the curly braces :D Going to add a linting rule so that never happens again ;P

But I don't get your first comment. Could you explain further?

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -130,6 +132,26 @@ export class UiControlClient extends FluentCommand {
     }
   }
 
+  private secretText: string | undefined = undefined;
+
+  override typeIn(text: string, { isSecret = false, secretMask = '****' } = {}): FluentFilters {
```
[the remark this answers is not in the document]
adi-wan-askui: The issue said that both is fine

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -109,16 +109,18 @@ export class UiControlClient extends FluentCommand {
 
   async exec(
     instruction: string,
-    customElementJson?: CustomElementJson[],
+    customElementJson: CustomElementJson[] = [],
   ): Promise<void> {
     let customElements: CustomElement[] = [];
     if (customElementJson !== undefined) {
```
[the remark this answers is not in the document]
adi-wan-askui: Good catch :) But actually we do not even need it anymore :)

packages/askui-nodejs/src/execution/execution-runtime.ts
programminx-askui: My fault. 

I thought when you are returning undefined in the function, that the reference of the promise is undefined instead of the promis is resolving with an undefined value. 

But this is a async function. So it wraps the return value in an promise

PR #22, opened 2022-06-23
AS-1010 Fix waitFor()

adi-wan-askui:
Follow-up:

- Adjust dsl template
- Remove `waits()` from ui controller

### DIFF
--- a/docs/docs/api/02-Commands/waitfor.md
+++ b/docs/docs/api/02-Commands/waitfor.md
@@ -6,4 +6,4 @@ custom_edit_url: null
 
 Waits for `<delayInMs>` ms, e.g., 1000 ms. The exact delay may be a little longer than `<delayInMs>` but never shorter than that.
 
-   * @param {number} delayInMs - The daily in ms to wait for.
+   * @param {number} delayInMs - The delay in ms to wait for.
--- a/docs/docs/api/02-Commands/waitfor.md
+++ b/docs/docs/api/02-Commands/waitfor.md
@@ -4,7 +4,6 @@ custom_edit_url: null
 ---
 # waitFor
 
-Waits for `<timeValue> <timeUnit>`, e.g., 10 seconds.
+Waits for `<delayInMs>` ms, e.g., 1000 ms. The exact delay may be a little longer than `<delayInMs>` but never shorter than that.
 
-   * @param {number} timeValue - A number of time unit, e.g., ms or min, to wait for.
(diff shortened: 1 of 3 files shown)


docs/docs/api/02-Commands/waitfor.md
```diff
@@ -4,7 +4,6 @@ custom_edit_url: null
 ---
 # waitFor
 
-Waits for `<timeValue> <timeUnit>`, e.g., 10 seconds.
+Waits for `<delayInMs>` ms, e.g., 1000 ms. The exact delay may be a little longer than `<delayInMs>` but never shorter than that.
 
-   * @param {number} timeValue - A number of time unit, e.g., ms or min, to wait for.
-   * @param {TimeUnit} timeUnit - A time unit, e.g., ms, s or min.
+   * @param {number} delayInMs - The daily in ms to wait for.
```
mlikasam-askui, line 9: delay*

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -130,6 +130,24 @@ export class UiControlClient extends FluentCommand {
...
+  // eslint-disable-next-line class-methods-use-this
+  waitFor(delayInMs: number): Executable {
+    return {
+      exec(): Promise<void> {
+        logger.debug(`Wait for ${delayInMs} ms`);
+        return new Promise((resolve) => { setTimeout(() => resolve(), delayInMs); });
+      },
+    };
+  }
+
```
mlikasam-askui, line 150: big like

mlikasam-askui (approved):

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -130,6 +130,24 @@ export class UiControlClient extends FluentCommand {
     }
   }
 
+  /**
+   * Waits for `<delayInMs>` ms, e.g., 1000 ms. The exact delay may be a little longer
```
mlikasam-askui, line 134: only question regarding,  The exact delay may be a little longer why is it the case

adi-wan-askui: Because you can never be a 100% sure when the event loop is going to pick this up again, only that it will be after > 1000 ms (see https://stackoverflow.com/questions/49811043/relationship-between-event-loop-libuv-and-v8-engine)

PR #33, opened 2022-06-24
docs: add troubleshooting for macos

adi-wan-askui:


mlikasam-askui:
from my perspective, this section should also be included in the doc for version 0.2.0.

mlikasam-askui (approved):
Nice writing skills :+1:

programminx-askui (approved):

adi-wan-askui:
> from my perspective, this section should also be included in the doc for version 0.2.0.

I thought so, too, at first. But this actually breaks links in the logs if we don't release a new lib version as well. So I would wait for next release. Let's release as soon as possible.

PR #35, opened 2022-06-27
feat(askui-nodejs): allow tracking time to value using install timest…

adi-wan-askui:
…amp (#AS-1070)

### DIFF
--- a/packages/askui-nodejs/bin/askui-postinstall
+++ b/packages/askui-nodejs/bin/askui-postinstall
@@ -1,12 +1,35 @@
 #!/usr/bin/env node
 
-const { InstallationTimestamp } = require('../dist/cjs/utils/analytics/installation-timestamp');
+const { access } = require('fs/promises');
+const { join } = require('path');
 
-(() => {
+async function shouldBeExecuted() {
   try {
+    await access(join(__dirname, '../.no-askui-postinstall'));
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      return true;
+    }
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+  return false;
+}
+
+async function createInstallationTimestamp() {
+  try {
+    const { InstallationTimestamp } = require('../dist/cjs/utils/analytics/installation-timestamp');
     InstallationTimestamp.create();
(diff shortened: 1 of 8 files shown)


packages/askui-nodejs/src/utils/analytics/installation-timestamp.ts
```diff
@@ -0,0 +1,48 @@
...
+    return InstallationTimestamp.value;
+  }
+
+  private static async getFromFile(): Promise<Date> {
+    return new Promise((resolve, reject) => {
+      fs.readFile(InstallationTimestamp.fileName, 'utf-8', (err, isoDateStr) => {
+        if (err) {
+          reject(new InstallationTimestampGetError(err));
+        } else {
+          resolve(new Date(isoDateStr));
```
mlikasam-askui, line 43: should we also check if isoDateStr is valid?

packages/askui-nodejs/bin/askui-postinstall
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env node
+
+const { access } = require('fs/promises');
+const { join } = require('path');
+
+async function shouldBeExecuted() {
+  try {
+    await access(join(__dirname, '../.no-askui-postinstall'));
```
mlikasam-askui, line 8: is there a specific reason why "access" is used here and not "exists"?

adi-wan-askui: `exists` is marked for deprecation and should generally not be used, see Node.js docs

packages/askui-nodejs/src/utils/analytics/installation-timestamp.ts
adi-wan-askui: It doesn't make sense from my perspective. It is unlikely that a user is going to mess with the contents of that file except people trying to misuse our api but we cannot prevent them from sending us bad data at this stage anyway. Let's say someone really changed it by accident, then, If the string was really not interpretable as a date, an error would be raised that would be caught and we would not get an installation timestamp.

mlikasam-askui (approved):

PR #41, opened 2022-07-05
fix(askui-nodejs): postinstall hook on windows

adi-wan-askui:


programminx-askui (approved):

packages/askui-nodejs/package.json
```diff
@@ -14,7 +14,7 @@
         "lint": "eslint --cache --fix --max-warnings 0 \"./**/*.{js,ts}\"",
         "lint:staged": "lint-staged",
         "test": "jest '/src'",
-        "postinstall": "$npm_package_bin_askui-postinstall"
+        "postinstall": "node -e \"require('./bin/askui-postinstall')\""
```
mlikasam-askui, line 17: just quick question, can'T this be replaced with npx askui-postinstall ?

adi-wan-askui: Nope, as `askui-postinstall` would not be found in this case

mlikasam-askui: thanks for the responce

PR #45, opened 2022-07-13
As 1137 migrate from tenant to workspace  model

mlikasam-askui:


packages/askui-nodejs/src/execution/inference-client.ts
```diff
@@ -8,10 +8,22 @@ import { resizeBase64ImageWithSameRatio } from '../utils/transformations';
...
+    readonly workspaceId?: string,
+    public apiVersion = 'v2',
+  ) {
+    this.url = workspaceId ? urljoin(
+      this.baseUrl,
+      'api',
+      this.apiVersion,
+      'workspaces',
+      workspaceId,
+    ) : urljoin(this.baseUrl, 'api', this.apiVersion);
```
adi-wan-askui, line 25: To make this code a little clearer, can we do?
```suggestion
    const versionedBaseUrl = urljoin(this.baseUrl, 'api', this.apiVersion);
    this.url = workspaceId ? urljoin(versionedBaseUrl, 'workspaces', workspaceId) : versionedBaseUrl;
```

packages/askui-nodejs/src/utils/http/credentials.ts
```diff
@@ -1,6 +1,5 @@
 export interface CredentialArgs {
   workspaceId: string,
```
adi-wan-askui, line 2: Looks like we don't even need the `workspaceId` in here. Can we remove it? Also from construction?

adi-wan-askui (approved):
Great job!

packages/askui-nodejs/src/utils/http/credentials.ts
mlikasam-askui: you mean inside class Credentials? 
the problem is, we need it only when the token Is defined and it's a part of the Credential. 
maybe we should encode it inside the token string, then it might be removed.

adi-wan-askui: you are right, I mean inside the class where we now can simply pass the token instead of credentials to the constructor.

packages/askui-nodejs/src/utils/http/credentials.ts
```diff
@@ -5,14 +5,14 @@ export interface CredentialArgs {
 
 export class Credentials {
   constructor(
-    private credentials: CredentialArgs,
```
adi-wan-askui, line 9: The interface should be moved to an extra file. They don't belong together anymore. The `CredentialArgs` is now an interface for the public api of the `UiControlClient` so I would move it close to that.

adi-wan-askui (approved):

adi-wan-askui (approved):

PR #72, opened 2022-09-01
Fix up Getting Started docs

adi-wan-askui:
So that installation of dependency does not run into conflicts

### DIFF
--- a/docs/docs/general/03-Best Practice/configure_test_framework.md
+++ b/docs/docs/general/03-Best Practice/configure_test_framework.md
@@ -4,32 +4,23 @@ custom_edit_url: null
 
 # Configure Test Frameworks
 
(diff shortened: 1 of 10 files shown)


docs/docs/general/03-Best Practice/configure_test_framework.md
```diff
@@ -4,32 +4,23 @@ custom_edit_url: null
...
-const config: Config.InitialOptions = {
+export default {
   preset: 'ts-jest',
   testEnvironment: 'node',
   setupFilesAfterEnv: ['./helper/jest.setup.ts'],
-  sandboxInjectedGlobals: [
-    'Math'
-  ]
-  
+  sandboxInjectedGlobals: ['Math']
```
programminx-askui, line 18: Do we still need Math? We've migrated away from jimp

programminx-askui: We should create a follow up ticket.

programminx-askui (approved):

PR #112, opened 2022-09-22
docs: fix wrong link in Getting Started to askui User Portal signup

JohannesDienst-askui:
I did not realise that _npm run start_ compiles even if there are wrong links.

But _npm run build_ fails with that.


adi-wan-askui (approved):
Nice catch :)

PR #171, opened 2022-10-20
docs: remove redirect from landing page; Correct GTM number for docs

JohannesDienst-askui:


docs/src/pages/index.js
```diff
@@ -22,7 +24,14 @@ function HomepageHeader() {
   );
 }
 
-import {Redirect} from '@docusaurus/router';
 export default function Home() {
-    return <Redirect to="/docs/general/Introduction/why-askui" />;
+  const { siteConfig } = useDocusaurusContext();
+  return (
+    <Layout>
```
thinh-askui, line 30: Nice catch! The four icons in the main part seems a bit to big for me?

docs/docusaurus.config.js
```diff
@@ -108,6 +108,14 @@ const config = {
                 label: 'Imprint',
                 href: 'https://www.askui.com/imprint',
               },
+              {
```
thinh-askui, line 111: Do we also need to add the script for google tag manager?

thinh-askui (commented):
Hey, nice catches again:) I just have two questions

docs/docusaurus.config.js
JohannesDienst-askui: Google recommends it as it would not work without it when a user has JavaScript disabled.

docs/src/pages/index.js
thinh-askui: Yes, looks better now

thinh-askui (approved):

PR #172, opened 2022-10-20
As 1405 get command frontend

mlikasam-askui:


docs/docs/api/01-API/table-of-contents.md
```diff
@@ -230,3 +230,9 @@ displayed_sidebar: apiSidebar
 
 -  [notExists](../checks/notexists)
 
+## Getters
+
+-  [getAll](../getters/getall)
+
+-  [get](../getters/get)
```
programminx-askui, line 237: @mlikasam-askui 

I thought you've changed the naming of the folders?

mlikasam-askui: it was not needed

PR #188, opened 2022-10-27
docs: upgrade npx askui init to take workspaceId and accessToken with…

JohannesDienst-askui:
Upgrade the command `npx askui init` to accept the `workspaceId` and the `accessToken` as options:

You can now do this:
```
npx askui init -w 123123123 -a 99999999 
npx askui init --workspace-id 123123123 --access-token 99999999
```

It then replaces the strings `<your workspace id>` `<your access token>` in the file `test/helper/jest.setup.ts`

We can provide this command in our user portal and setup becomes much easier!

---------------------------------------------------------------------------------------------------------------------
Upgrade the test case of the example project to do the following:

1. annotateInteractively() -> So the user immediately sees something and can explore the annotation
2. move the mouse to the upper left corner -> User experiences mouse movement
3. the mouse moves back and clicks on the text in the test file

I hope this accomplishes two things:

1. The user gets feedback if the tool is running via annotation and later with watchable movement
2. The mouse movement should give an immediate 'aha'-effect because it is real movement and no jumping around

---------------------------------------------------------------------------------------------------------------------

### DIFF
--- a/packages/askui-nodejs/example_projects_templates/typescript_jest/test/helper/jest.setup.ts
+++ b/packages/askui-nodejs/example_projects_templates/typescript_jest/test/helper/jest.setup.ts
@@ -1,4 +1,4 @@
-import { UiControlClient, UiController } from 'askui';
+import { UiControlClient, UiController } from 'askui'; // eslint-disable-line import/no-extraneous-dependencies
 
 // Server for controlling the operating system
 let uiController: UiController;
@@ -20,7 +20,12 @@ beforeAll(async () => {
 
   await uiController.start();
 
-  aui = await UiControlClient.build();
+  aui = await UiControlClient.build({
+    credentials: {
+      workspaceId: '<your workspace id>',
+      token: '<your access token>',
+    },
+  });
 
   await aui.connect();
 });
--- a/packages/askui-nodejs/example_projects_templates/typescript_jest/test/helper/jest.setup.ts
+++ b/packages/askui-nodejs/example_projects_templates/typescript_jest/test/helper/jest.setup.ts
@@ -1,4 +1,4 @@
-import { UiControlClient, UiController } from 'askui'; // eslint-disable-line import/no-extraneous-dependencies
+import { UiControlClient, UiController } from 'askui';
 
 // Server for controlling the operating system
 let uiController: UiController;
--- a/packages/askui-nodejs/example_projects_templates/typescript_jest/test/my-first-askui-test-suite.test.ts
+++ b/packages/askui-nodejs/example_projects_templates/typescript_jest/test/my-first-askui-test-suite.test.ts
@@ -9,6 +9,7 @@ describe('jest with askui', () => {
     await aui.moveMouse(0, 0).exec();
     await aui
       .click()
+      .text()
       .withText('Click on this text right here!')
       .exec();
   });
--- a/packages/askui-nodejs/example_projects_templates/typescript_jest/test/my-first-askui-test-suite.test.ts
+++ b/packages/askui-nodejs/example_projects_templates/typescript_jest/test/my-first-askui-test-suite.test.ts
@@ -2,9 +2,14 @@ import { aui } from './helper/jest.setup';
 
 describe('jest with askui', () => {
   it('should click on text', async () => {
+    // Run this to see what askui annotates
+    await aui.annotateInteractively();
+
+    // TODO Validate this works
+    await aui.moveMouse(0, 0).exec();
     await aui
       .click()
-      .text()
+      .withText('Click on this text right here!')
       .exec();
   });
 });
--- a/packages/askui-nodejs/example_projects_templates/typescript_jest/test/my-first-askui-test-suite.test.ts
+++ b/packages/askui-nodejs/example_projects_templates/typescript_jest/test/my-first-askui-test-suite.test.ts
@@ -5,7 +5,6 @@ describe('jest with askui', () => {
     // Run this to see what askui annotates
     await aui.annotateInteractively();
 
-    // TODO Validate this works
     await aui.moveMouse(0, 0).exec();
     await aui
       .click()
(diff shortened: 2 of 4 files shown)


packages/askui-nodejs/example_projects_templates/typescript_jest/test/helper/jest.setup.ts
```diff
@@ -1,4 +1,4 @@
-import { UiControlClient, UiController } from 'askui';
+import { UiControlClient, UiController } from 'askui'; // eslint-disable-line import/no-extraneous-dependencies
```
adi-wan-askui, line 1: We should ignore this rule for the whole `example_projects_templates`. I am a little surprised that this linting error only arose now... are your sure that this file is even linted?

packages/askui-nodejs/example_projects_templates/typescript_jest/test/my-first-askui-test-suite.test.ts
```diff
@@ -2,9 +2,14 @@ import { aui } from './helper/jest.setup';
...
   it('should click on text', async () => {
+    // Run this to see what askui annotates
+    await aui.annotateInteractively();
+
+    // TODO Validate this works
+    await aui.moveMouse(0, 0).exec();
     await aui
       .click()
-      .text()
+      .withText('Click on this text right here!')
```
adi-wan-askui, line 12:
```suggestion
      .text()
      .withText('Click on this text right here!')
```

packages/askui-nodejs/src/lib/copy-example-project.ts
```diff
@@ -1,24 +1,53 @@
...
 export function init(argv: string[]): Command {
   const args = argv || process.argv;
   const program = createProgram();
+
   program
     .command('init')
     .description('creates a typescript example project')
-    .action(() => { copyExampleProject(); });
+    .option('-w, --workspace-id <value>', 'the workspace id')
+    .option('-a, --access-token <value>', 'the access token')
```
adi-wan-askui, line 47:
```suggestion
    .option('-w, --workspace-id <value>', 'a workspace id')
    .option('-a, --access-token <value>', 'an access token for the workspace with the id')
```

packages/askui-nodejs/src/lib/copy-example-project.ts
```diff
@@ -1,24 +1,53 @@
...
 };
-function copyExampleProject() {
+
+async function replaceStringInFile(filePath: string, replace: string, replacement: string) {
+  try {
+    const data = await fs.readFile(filePath, 'utf8');
+    const result = data.replace(replace, replacement);
+    await fs.writeFile(filePath, result, 'utf8');
+  } catch (error: unknown) {
+    logger.error(`Could not replace '${replace}' with ${replacement} in file ${path}`);
```
adi-wan-askui, line 19: `${replacement}` should also be surrounded by quotation marks

packages/askui-nodejs/src/lib/copy-example-project.ts
```diff
@@ -1,24 +1,53 @@
...
+    if (error instanceof Error) {
+      logger.error(error.message);
+    }
```
adi-wan-askui, line 22: Can `error` be something else than an `Error`? If not (and I guess not), I would like a cast better as it conveys that meaning, i.e.,
```suggestion
    logger.error((error as Error).message);
```

adi-wan-askui (changes requested):
Nice improvement to the setup experience :)

packages/askui-nodejs/src/lib/copy-example-project.ts
JohannesDienst-askui: I read up on this and for TypeScript it is always _unknown_ (See https://stackoverflow.com/questions/42618089/how-do-you-use-typed-errors-in-async-catch)

So I rewrote it with a cast that checks for the message property and if it is there it will log the message.

packages/askui-nodejs/example_projects_templates/typescript_jest/test/helper/jest.setup.ts
JohannesDienst-askui: If I run `npm run lint` it fails for this line 😞 

I will add an override block for the _example_project_templates_ folder

packages/askui-nodejs/src/lib/copy-example-project.ts
JohannesDienst-askui: And for the _path_ also! I was sloppy there...

packages/askui-nodejs/src/lib/copy-example-project.ts
JohannesDienst-askui: Makes more sense to add the hint _for the workspace with the id_. Nice addition!

For me that was obvious 😄

packages/askui-nodejs/example_projects_templates/typescript_jest/test/my-first-askui-test-suite.test.ts
JohannesDienst-askui: @adi-wan-askui Changed and pushed!

packages/askui-nodejs/src/lib/copy-example-project.ts
JohannesDienst-askui: @adi-wan-askui Changed and pushed!

packages/askui-nodejs/src/lib/copy-example-project.ts
JohannesDienst-askui: @adi-wan-askui Changed and pushed!

packages/askui-nodejs/src/lib/copy-example-project.ts
JohannesDienst-askui: @adi-wan-askui Changed and pushed!

adi-wan-askui (approved):
Trust you to fix the last TS issue :) Nice work!

PR #199, opened 2022-11-04
docs: Lots of small improvements

JohannesDienst-askui:
### Update to docusaurus 2.2.0
No more to see here 😉 

### Getting Started
* Prerequisite gets an info admonition
* Hide explanation of dependencies in a detail block: This is only possible with HTML unfortunately :( Docusaurus own documentation uses HTML there!

### Spell-Checking for Docs
Add spellchecker-cli to preCommit hook.
Decided to go with only scanning the _next_ Docs as correcting all the errors in _versioned_docs_ would take me a day or so and also the scanning takes too much time (a few minutes)


.husky/pre-commit
```diff
@@ -5,5 +5,8 @@ npx is-ci && exit 0
 npm run lint:staged
 npm test
 
-cd ./examples/askui-nodejs-base-example || exit
+cd ./docs || exit
+npm run spellcheck-docs
```
programminx-askui, line 9: like :+1:

docs/docs/api/02-Commands/swipe.md
```diff
@@ -5,5 +5,5 @@ displayed_sidebar: apiSidebar
 
 Swipe an element in the x and y direction
 
-   * @param {number} x_offset - A x direction. positive an negative values are accepted
-   * @param {number} y_offset - A y direction. positive an negative values are accepted
+   * @param {number} x_offset - A x direction. positive and negative values are accepted
+   * @param {number} y_offset - A y direction. positive and negative values are accepted
```
programminx-askui, line 9: Did you changed all the spell correction inside the askui-inference repro, too? Otherwise they get overwritten

docs/docs/general/07-Troubleshooting/index.md
```diff
@@ -6,8 +6,8 @@ challenge after you have read through the pages, please drop us a note. We are h
 ## Pages
 * [Deprecated Endpoints](deprecated-endpoints.md)
 * [ipv6](ipv6.md)
-* [linux](linux.md)
-* [mac-os](mac-os.md)
+* [Linux](linux.md)
+* [Mac OS](mac-os.md)
```
programminx-askui, line 10:
```suggestion
* [macOS](mac-os.md)
```

See https://en.wikipedia.org/wiki/MacOS

docs/docs/api/02-Commands/swipe.md
JohannesDienst-askui: Yes: See https://github.com/askui/askui-inference/pull/73

On another note. Is the documentation automatically overridden? I want to add more docs (gifs, extended examples to each command). I do not think inference would be the right place to hold it, or is it?

programminx-askui: The API folder gets overwritten by any changes of the dsl in askui-inference. --> The changes should be adapted inside the askui-inference. 

The documentation is used inside the esdoc (inline code documentation) and the docs (docusaurus). If you add some gifs, we have to check if the in-line doc can display them too. 

We should have a talk

programminx-askui (approved):

PR #215, opened 2022-11-17
docs: add gifs with an example for relational selectors under API

JohannesDienst-askui:


thinh-askui (approved):
Cool idea - looking really good :)

PR #219, opened 2022-11-22
docs: fix missing function in Table of Contents

HaramChoi-askui:
added 'withTextRegex' to Table of Contents of version>0.2.2


thinh-askui (changes requested):
Nice Catch! You added all changes to the `docs/versioned_docs`. 
Need to change it in the `docs/docs` folder too. A new release will use this version

PR #268, opened 2023-01-20
fix(template): rm color descriptions (#AS-1711)

mlikasam-askui:


programminx-askui:
This could also be fixed inside the askui-inference backend

mlikasam-askui:
@programminx-askui ? i though we need to remove everything related to colors inside our code base

PR #273, opened 2023-01-26
fix: add mouse commands

mlikasam-askui:


docs/docusaurus.config.js
```diff
@@ -10,146 +10,142 @@ const tagline = 'Humanizing UI Automation';
...
 const isProd = process.env.NODE_ENV === 'production';
 if (!isProd) {
-  headScript = '/scripts/isNotProd.js';
+    headScript = '/scripts/isNotProd.js';
 }
 ////////////////////////////////////////////////////
 
 /** @type {import('@docusaurus/types').Config} */
 const config = {
-  title: `askui - ${tagline}`,
```
JohannesDienst-askui, line 19: Can you format it the way it was before, so that we only have the fix for the missing page?

JohannesDienst-askui (changes requested):

JohannesDienst-askui (approved):

PR #274, opened 2023-01-27
Give users usage warnings in time

adi-wan-askui:


packages/askui-nodejs/src/core/inference-response/inference-response.ts
```diff
@@ -1,41 +1,34 @@
...
-      image: string,
-    ) => Annotation
-      .fromJson({ image, detected_elements: data.detected_elements }, resizeRatio),
-    COMMANDS: (data: ControlCommand, resizeRatio: number) => ControlCommand
-      .fromJson(data, resizeRatio),
-  };
-}
-
-interface Models {
-  DETECTED_ELEMENTS: CallableFunction
```
mlikasam-askui, line 39: can you explain why you changed it?

packages/askui-nodejs/src/utils/http/http-client-got.ts
```diff
@@ -15,45 +15,68 @@ export class HttpClientGot {
     readonly token?: string,
     readonly customHeaders?: Record<string, string>,
     private readonly cookies: Record<string, string> = {},
-    readonly proxyAgents?: { http: http.Agent, https: https.Agent },
+    readonly proxyAgents?: { http: http.Agent; https: https.Agent },
```
programminx-askui, line 18: Why did you switched to a semicolon?

packages/askui-nodejs/src/utils/http/http-client-got.ts
```diff
@@ -15,45 +15,68 @@ export class HttpClientGot {
...
+      responseType: 'json',
+      throwHttpErrors: false,
+    });
+    const { body, statusCode, headers } = await this.askuiGot.post<T>(
+      url,
+      options,
+    );
     if (headers['deprecation'] !== undefined) {
       logger.warn(headers['deprecation']);
     }
```
programminx-askui, line 65: @adi-wan-askui  Did you saw this? We already check here the header.

programminx-askui: We could move this to the `inference-client.ts` too.

programminx-askui (commented):
Only two small remarks

packages/askui-nodejs/src/core/inference-response/inference-response.ts
adi-wan-askui: Typing was incorrect in some places. Fixing that lead to the typing here being incorrect etc. so I changed it as well...

packages/askui-nodejs/src/utils/http/http-client-got.ts
adi-wan-askui: I decided against it because the `deprecation` header is a generic header that is generally applicable, see https://greenbytes.de/tech/webdav/draft-ietf-httpapi-deprecation-header-latest.html (although we currently don't use it as specified there).

packages/askui-nodejs/src/utils/http/http-client-got.ts
adi-wan-askui: Was my formatter... Currently, we do not enforce this automatically so I would keep it this way until we do...

packages/askui-nodejs/src/core/inference-response/inference-response.ts
mlikasam-askui: my question was mainly regarding introducing the if conditions.

adi-wan-askui: As easy to change and read as the map option so not degregation in code quality  but easier typing...

programminx-askui (approved):

PR #292, opened 2023-02-09
Rename speed-up-execution to recommended practices and remodel its content

JohannesDienst-askui:


thinh-askui (approved):
Sorry, was a bit slow. Nice changes

PR #326, opened 2023-03-08
docs: Terminology updates for whole docs -> Next-version

JohannesDienst-askui:


HaramChoi-askui (changes requested):
- Line no. 158: a instruction -> an instruction

PR #328, opened 2023-03-09
fix(ui-control-client): fix type and typeIn secret text bug

mlikasam-askui:
Fix: type and typeIn secret text bug


adi-wan-askui (changes requested):
Can we reset it after command execution inside `fluentCommandExecutor()` instead immediately after it is used? Seems clearer to me.

adi-wan-askui (approved):

PR #329, opened 2023-03-09
docs: Fix trailing slash problem when reloading the API docs Table of Contents page

JohannesDienst-askui:


JohannesDienst-askui:
@adi-wan-askui This may brake in production in an unforeseen way as Github pages is different then localhost.

I tested with a `production` build on localhost and there I could not find broken links 🤞

adi-wan-askui (approved):

PR #395, opened 2023-04-29
docs: add gif examples to the api docs

HaramChoi-askui:
Added animated gif examples to the API docs.
I just realized that I missed some gifs, but will add them soon.


docs/docs/api/03-Element-Descriptions/switch.md
```diff
@@ -6,3 +6,9 @@ displayed_sidebar: apiSidebar
 
 Filters for a UI element 'switch'.
 
+**Example:**
+```typescript
+await aui.click().swith().rightOf()
```
JohannesDienst-askui, line 11: Not sure if that is in askui-inference also. It should be `switch`

JohannesDienst-askui (changes requested):
This looks great! I like how professional the gifs look 🕺 

Does it make sense to gifs for `text` and `textfield`?

Also is there a PR at askui-inference with the changes?

JohannesDienst-askui (approved):

PR #405, opened 2023-05-08
feat: add copy to clipboard icon to the annotation

mlikasam-askui:
- fix copy2Code  function
- add copy2Clipboard icon


packages/askui-nodejs/src/execution/inference-client.ts
```diff
@@ -9,7 +9,7 @@ import { IsImageRequired } from './is-image-required-interface';
 import { InferenceResponseError } from './inference-response-error';
 import { DetectedElement } from '../core/model/annotation-result/detected-element';
 import { ConfigurationError } from './config-error';
-import { InferenceResponseBody } from '@/core/inference-response/inference-response';
+import { InferenceResponseBody } from '../core/inference-response/inference-response';
```
programminx-askui, line 12: ?

@adi-wan-askui is the `@` important?

adi-wan-askui: Nope, it is simply a path alias so that you don't have to do `../../..`. I think not using it here is fine as the path is still clear. But I would start using it if you have to got up more than 2 folders. In doubt, always use it. Cannot hurt as it makes the path clearer.

PR #436, opened 2023-05-30
docs: short guide for get()

JohannesDienst-askui:
Add a short guide on how to use `get()` to extract information out of the UI.

### DIFF
--- a/docs/docs/general/03-Guides/extract-information.md
+++ b/docs/docs/general/03-Guides/extract-information.md
@@ -1,0 +1,76 @@
+---
+sidebar_position: 7
+---
+
+# Extract Information with get()
+In this guide you will learn how to extract elements and their values out of your User Interface.
+
+We will also cover on how to use the extracted information in your code, so you can introduce real interactivity into your workflow.
+
+## Basic Usage
+The [get()](../../api/06-Getters/get.md)-Instruction returns an array of all the elements it found. It looks like this example:
+
+```text 
+ console output: [
+  DetectedElement {
+     name: 'TEXT',
+     text: 'Sign In',
+     bndbox: BoundingBox {
+        xmin: 1128.2720982142857,
+        ymin: 160.21332310267857,
+        xmax: 1178.8204241071428,
+        ymax: 180.83512834821428
+     }
+   },
+   ...
+ ]
+```
+
+Usually you do not want every element from your UI returned. So you pair it with an element-selector:
+
+```typescript
+const textFieldText = 
+  await aui
+    .get()
+    .text()
+    .in()
+    .textfield()
+    .exec();
+```
+
+
+
+```text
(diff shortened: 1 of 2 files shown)


docs/docs/general/03-Guides/extract-information.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Extract the Value
+You can extract the value by reading the `text`-property of the `DetectedElement`.
+
+**Notice: `textFieldText` is an array. You have to iterate over it or use an index to access the `DetectedElement`.**
+
+```typescript
+// Access by index
+// textFieldTextValue gets the value 'Search'
+const textFieldTextValue = textfieldText[0].text;
```
adi-wan-askui, line 62: You should make sure that testfieldText[0] is actually set (and correctly cased ;P), i.e., the array is not empty, shouldn't you? E.g.,
```typescript
const textFieldTextValue = textFieldText[0]?.text
```
or even
```typescript
const textFieldTextValue = textFieldText?.[0]?.text
```

adi-wan-askui (changes requested):
Nice :)

docs/docs/general/03-Guides/extract-information.md
JohannesDienst-askui: What a nasty copy&paste error from me 😢  Thanks for catching it!

Also will include the suggestion with `?`

In my opinion the second `?` is not needed as `text` is always defined in this case. Any objections?

docs/docs/general/03-Guides/extract-information.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+
+```text
+[
+  DetectedElement {
+    name: 'text',
+    text: 'Search',
+    bndbox: BoundingBox { xmin: 488, ymin: 138, xmax: 548, ymax: 164 },
+    colors: [ '', '', '' ]
```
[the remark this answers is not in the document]
JohannesDienst-askui: In version `0.10.1` the output is still like this. How will it change @mlikasam-askui ?

docs/docs/general/03-Guides/extract-information.md
```diff
@@ -59,7 +59,7 @@ You can extract the value by reading the `text`-property of the `DetectedElement
 ```typescript
 // Access by index
 // textFieldTextValue gets the value 'Search'
-const textFieldTextValue = textfieldText[0].text;
+const textFieldTextValue = textFieldText?[0]?.text;
```
adi-wan-askui, line 62: Is this right? I think it is:
```suggestion
const textFieldTextValue = textFieldText?.[0]?.text;
```

adi-wan-askui (approved):
Feel free to resolve the remarks and merge :)

PR #440, opened 2023-06-06
docs: try to fix error with CookieConsent not loading YouTube embeds

JohannesDienst-askui:


adi-wan-askui (approved):
Should also fix the issue with Youtube videos not being displayed in Safari (`&embedded`) because they are loaded from different origin.

PR #452, opened 2023-06-13
fix(packages/askui-nodejs): fix that text and index params are optional

adi-wan-askui:
- accepted undefined instead of being optional

### DIFF
--- a/packages/askui-nodejs/src/execution/dsl.ts
+++ b/packages/askui-nodejs/src/execution/dsl.ts
@@ -283,7 +283,7 @@ export class FluentFilters extends FluentBase {
    * @return {FluentFiltersOrRelations}
    */
   text(
-    text: string | undefined,
+    text?: string,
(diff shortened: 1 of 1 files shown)


adi-wan-askui:
@mlikasam-askui Please, adjust the generator.

packages/askui-nodejs/src/execution/dsl.ts
```diff
@@ -807,8 +805,7 @@ export class FluentFiltersOrRelations extends FluentFilters {
    * @return {FluentFilters}
    */
   below(
-    optionalIndex: number | undefined
-    = 0,
+    optionalIndex = 0,
   ): FluentFilters {
```
mlikasam-askui, line 809: below(undefined) is valid too , and it's the inference response when we request text below text to fluent api => text(undefined).below(undefined).text(undefined)

PR #459, opened 2023-06-21
feat(askui-nodejs): retry failed requests to inference

adi-wan-askui:


packages/askui-nodejs/src/utils/http/http-client-got.ts
```diff
@@ -11,14 +17,83 @@ export class HttpClientGot {
...
+
+          if (
+            error.response !== undefined
+            && error.response.headers['retry-after'] === undefined
+          ) {
+            return Math.min(
+              1000 * 2 ** (attemptCount - 1) + Math.random() * 100,
+              Number.MAX_SAFE_INTEGER,
+            );
+          }
```
mlikasam-askui, line 72: should we log a warning after each retry?  
and can this cause a timeout error?

mlikasam-askui (approved):
Nice Job :+1: ,
I have only a small question

packages/askui-nodejs/src/utils/http/http-client-got.ts
adi-wan-askui: Thx for pointing out the missing logging. I add it but on a debug level as I would like to hide that as long as the last request succeeds. 

No timeout error in the integration tests but could happen potentially which I think is fine.

PR #470, opened 2023-07-03
docs: update all pages that use code like text().withText() to the op…

JohannesDienst-askui:
…tional parameter version

### DIFF
--- a/docs/docs/api/03-Element-Descriptions/text.md
+++ b/docs/docs/api/03-Element-Descriptions/text.md
@@ -17,10 +17,10 @@ await aui.click().text().exec();
 await aui.click().text('Username').exec();
 
 // Matching with an exact text
-await aui.click().text().withExactText('Username').exec();
+await aui.click().text('Username').exec();
 
 // Matching with a regex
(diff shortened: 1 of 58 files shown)


docs/docs/api/03-Element-Descriptions/text.md
HaramChoi-askui, line 1: The `text()` didn't work with regex in my test.
This works (without `text()`):
`await aui.moveMouseTo().withTextRegex('\b[Ss]\w+').exec()`

Also, the `withExactText()` remains necessary in my test. Because the `text()` approximates the pattern.

So I guess we have to keep this part of the docs for text descriptors.

(tested with the version 0.10.5)

HaramChoi-askui (changes requested):
Amazing, thanks for applying the change 👍 

But I guess we have to keep the part for `withTextRegex()` and `withExactText()`.

docs/docs/api/03-Element-Descriptions/text.md
JohannesDienst-askui: Great catch! This is a mistake from my side as I did not want to change this part of the code.

PR #477, opened 2023-07-05
docs: add exports for Annotation and DetectedElements so a Reporter can use them

JohannesDienst-askui:


adi-wan-askui (approved):

adi-wan-askui:
This is no change to docs or is it? 😅

PR #481, opened 2023-07-06
deprecate annotation level

adi-wan-askui:
- feat(packages/askui-nodejs)!: deprecate creating annotation file after instruc. (incl. config)
- fix(packages/askui-nodejs): fix auto-completion for reporter in UiControlClient's ClientArgs

### DIFF
--- a/docs/docs/api/08-Configuration/askui-ui-control-client.md
+++ b/docs/docs/api/08-Configuration/askui-ui-control-client.md
@@ -50,29 +50,6 @@ ___
   The side length of the target image to resize to in px. Your screenshot image will be resized with the original aspect ratio, and the lengths image side will be equal to this number. This can be used to reduce the inference time by reducing the request size in case of a bad internet connection. But it can cause a decrease in the prediction quality.
   The resizing will be skipped if it's undefined.
 
-
-___
-
-### annotationLevel
-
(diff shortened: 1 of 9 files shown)


JohannesDienst-askui:
This looks good 😃  Tested it locally.

Two questions just for my understand:

* `AnnotationLevel` gets deprecated because we want the specific configured reporter to handle this stuff
* Are we OK that some users will see a compile error because the `annotationLevel` is not a property for `UiControlClient` `ClientArgs` any more? I mean it is a quick fix to remove it 😉

adi-wan-askui:
> This looks good 😃  Tested it locally.
> 
> 
> 
> Two questions just for my understand:
> 
> 
> 
> * `AnnotationLevel` gets deprecated because we want the specific configured reporter to handle this stuff
> 
> * Are we OK that some users will see a compile error because the `annotationLevel` is not a property for `UiControlClient` `ClientArgs` any more? I mean it is a quick fix to remove it 😉  

With regards to the first question: yes, exactly. Basically, the annotation functionality is currently duplicated inside the report. The reporter provides also more details. For example, which steps the annotation belongs to.

With regards to the second question: I implemented it as a hard deprecation because I don't think a lot of people use it and if they do, the fix is easy.

JohannesDienst-askui (approved):

PR #509, opened 2023-07-28
fix: fix user identifier (#CO-411)

mlikasam-askui:
**Jira**
CO-411


packages/askui-nodejs/src/utils/analytics/user-identifier.ts
```diff
@@ -4,6 +4,10 @@ import { UserIdentifierInterface } from './user-identifier-interface';
 export class UserIdentifier implements UserIdentifierInterface {
   // eslint-disable-next-line class-methods-use-this
   async userId(): Promise<string> {
-    return machineId();
+    try {
+      return await machineId();
+    } catch (error) {
+      return Promise.resolve('undefined');
```
programminx-askui, line 10: can we add an debug log?

programminx-askui: log.debug("Can't retrieve user id!", error)

programminx-askui (approved):

programminx-askui (approved):

PR #537, opened 2023-08-16
Askui-Lib: Annotation renders a provided set of elements only

mlikasam-askui:
- add elements as optional input to annoation command, so it can be used as a debug tool

**Jira** CO-416


packages/askui-nodejs/src/core/model/annotation-result/annotation-interface.ts
```diff
@@ -1,8 +1,10 @@
 import { CustomElementJson } from '../custom-element-json';
+import { DetectedElement } from './detected-element';
 
 export interface AnnotationRequest {
   imagePath?: string,
+  elements?: DetectedElement[]
   outputPath?: string,
-  fileNamePrefix?:string,
+  fileNamePrefix?: string,
```
dom-askui, line 8: Automated formatting or you having an eye for details? ;) It's important to have an eye for details! Details make the difference between good enough and awesome!

dom-askui (approved):

packages/askui-nodejs/src/core/model/annotation-result/annotation-interface.ts
mlikasam-askui: auto-formated

PR #561, opened 2023-09-12
feat(ui-controller): add runtime option

mlikasam-askui:
- add runtime option to the ui-controller.

**TODO**
- Adjust the android tutorials. 

**JIRA**: ND


mlikasam-askui:
closed because it was not planned and we need to discuss it first.

packages/askui-nodejs/src/lib/ui-controller-args.ts
```diff
@@ -43,6 +43,7 @@ export interface UiControllerArgs {
   readonly logLevel?: LogLevels;
   readonly logFilePath?: string;
   readonly proxyAgents?: ProxyAgentArgs
+  readonly runtime?: 'desktop' | 'android'
```
dom-askui, line 46: That was the '-r  android' we used, right? 

I guess for now we have some limitations here. Later I would prefer a '--device <device_name>' option.

PR #562, opened 2023-09-26
feat(packages/askui-nodejs): allow passing multiple reporters to ui control client (#CL-457)

adi-wan-askui:
- ...instead of only one reporter
- make reporting asynchronous using promise chaining to not degrade / improve performance
- add/improve tests


packages/askui-nodejs/src/core/reporting/step-reporter.ts
```diff
@@ -4,68 +4,121 @@ import { Reporter } from './reporter';
...
+    return Promise.all(
+      this.reporters.map((reporter) => reporter.promiseQueue),
+    ).then(() => {});
+  }
 
-  get config(): Required<ReporterConfig> {
-    return this.reporter.config;
+  get currentStep(): DefaultStep | undefined {
+    return this._currentStep;
   }
```
mlikasam-askui, line 65: is this being used ?

mlikasam-askui (approved):
i have only a small question

packages/askui-nodejs/src/core/reporting/step-reporter.ts
adi-wan-askui: Yes, in the test of the corresponding class.

PR #573, opened 2023-10-27
feat: Interactive CLI

JohannesDienst-askui:
### Feature Description
Introduction of an interactive CLI where you can run `npx askui@latest init` and it asks as bunch of questions to set up AskUI in one go without the need to fumble with dependencies.

#### Functionality
The CLI asks about the following things which can also be directly given to the CLI with parameters:

- Platform (windows | macos | linux)
- Programming Language (Only TypeScript for now)
- Test Framework (jest | jasmine)
- Workspace ID
- Access Token
- Proxy Usage
- Creation of tsconfig

Depending on the answers it initializes an askui setup in the directory `askui_examples`, installs all necessary dependencies and populates the config files.

#### Docs

- The docs now contain the simplified `npm run askui` command everywhere instead of the `npx jest ...` one.
- The Getting Started Guides where updated, reflecting the different approaches for Windows, Linux and macOS

## How to Test

- Interactive CLI:
   `cd packages/askui-nodejs`
   `npm run build`
   Then you can execute the built version (see `dist` folder) by running the `bin/askui` executable in an empty folder for example like this:
   `./<path to bin>/askui`
- Docs: `cd docs` -> `npm run build` -> `npm run serve`


JohannesDienst-askui:
@premith-askui Please take a look.

packages/askui-nodejs/src/lib/interactive_cli/cli.ts
```diff
@@ -0,0 +1,149 @@
...
+    type: 'list',
+    name: 'operatingSystem',
+    message: 'Which operating system are you on?',
+    choices: ['windows', 'linux', 'macos'],
+  },
+  {
+    type: 'list',
+    name: 'progLanguage',
+    message: 'Which language do you prefer?',
+    choices: ['typescript'],
```
premith-askui, line 19: Do we need this or did you miss some choices here. like `javascript`

packages/askui-nodejs/src/lib/interactive_cli/cli.ts
```diff
@@ -0,0 +1,149 @@
...
+
+      if (missingOptions.length > 0) {
+        inquirer.prompt(missingOptions)
+          .then(async (answers) => {
+            Object.assign(userAnswers, answers);
+            await (new CreateExampleProject(userAnswers as CliOptions)).createExampleProject();
+          });
+      } else {
+        await (new CreateExampleProject(userAnswers as CliOptions)).createExampleProject();
+      }
```
premith-askui, line 146:
```suggestion
    const userAnswers = programInit.opts<CliOptions>();

    // Check which options are missing and prompt the user for them
    const missingOptions = questions.filter(entry => !(entry.name in userAnswers));

    if (missingOptions.length > 0) {
      inquirer.prompt(missingOptions).then(async (answers) => {
        await new CreateExampleProject({
          ...userAnswers,
          ...answers,
        }).createExampleProject();
      });
    } else {
      await new CreateExampleProject(userAnswers).createExampleProject();
    }

File: packages/askui-nodejs/src/lib/interactive_cli/cli.ts [hunk 3]
```diff
@@ -0,0 +1,149 @@
+import inquirer, { QuestionCollection } from 'inquirer';
+import { Command, Option, OptionValues } from 'commander';
+import { CreateExampleProject } from './create-example-project';
+import { CliOptions } from './cli-options-interface';
+
+const nonEmpty = (subject: string) => (input: string) => (input.trim().length > 0) || `${subject} is required.`;
+
+const questions: QuestionCollection = [
```
premith-askui, line 8: If we remove the type here we could simplify the code as suggested below

packages/askui-nodejs/src/lib/interactive_cli/cli.ts
JohannesDienst-askui: I missed that while I was copying it over from an old branch. Good catch!

packages/askui-nodejs/src/lib/interactive_cli/cli.ts
```diff
@@ -0,0 +1,149 @@
...
+  {
+    type: 'list',
+    name: 'operatingSystem',
+    message: 'Which operating system are you on?',
+    choices: ['windows', 'linux', 'macos'],
```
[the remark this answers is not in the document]
JohannesDienst-askui: Easily doable with `platform.process`. Could have used `os.platform()` also but that would have required an import.

packages/askui-nodejs/src/lib/interactive_cli/cli.ts
JohannesDienst-askui: Worked for me. Please retest 😃

premith-askui (approved):

PR #578, opened 2023-11-07
[Tools | Environment] Integration of Askui-Lib and RemoteDeviceController installation

mlikasam-askui:
- Integration of Askui-Lib and RemoteDeviceController installation.
- Change Askui-Lib Cli windows example project.
- Remove Operating System detection.

**Jira:** CO-538


JohannesDienst-askui (commented):
@mlikasam-askui Cool, you added the check for the Installation Directory 🥳 

Can you explain why you need the start of the UiController in `askui-helper.ts` if the Remote Device Controller is already installed and running? I deliberately took it out and created a separate template for Windows for now because of it.

I am missing something here and can not find it in the code.

mlikasam-askui:
> @mlikasam-askui Cool, you added the check for the Installation Directory 🥳
> 
> Can you explain why you need the start of the UiController in `askui-helper.ts` if the Remote Device Controller is already installed and running? I deliberately took it out and created a separate template for Windows for now because of it.
> 
> I am missing something here and can not find it in the code.

@JohannesDienst-askui, we need the UiController in askui-helper.ts for the following reasons:
1. Users do not need to manually start the Remote Device Controller each time before running the tests.
2. AskUI-Lib can still be used without the installer.
3. Users can select the screen and set the logFilePath with the help of the UiController class.

JohannesDienst-askui:
@mlikasam-askui Ah I see how it works now.

My understanding of the intended flow for a user was this:
1. Use the installer to install the Remote Device Controller
2. Start it and select the display
3. Create AskUI project

If this is not the intended flow we want our users to take, then I will approve the PR once https://docs.askui.com/docs/general/Getting%20Started/getting-started#download-remote-device-controller is adapted as it will be wrong after merge 😉

JohannesDienst-askui:
@mlikasam-askui Also how is the new CLI that comes with the installer for the Remote Device Controller solving this? We should make the user experience similar in my opinion.

JohannesDienst-askui:
@mlikasam-askui Would you move your last commit with the video enablement only in Jest to a separate PR?

We still have not discussed how to proceed with the other commits 😉 .

JohannesDienst-askui:
Will close this as it is outdated and Jasmine has been removed for now.

PR #583, opened 2023-12-11
Replace local search with lunr search

JohannesDienst-askui:


docs/docusaurus.config.js
```diff
@@ -43,6 +43,26 @@ const config = {
...
+          '/docs/general/**/*',
+          '/docs/next/api/API/*',
+          '/docs/next/general/*',
+          '/docs/next/general/**/*',
+          '/docs/0.11.6/api/API/*',
+          '/docs/0.11.6/general/*',
+          '/docs/0.11.6/general/**/*',
+          '/docs/0.12.2/api/API/*',
+          '/docs/0.12.2/general/*',
+          '/docs/0.12.2/general/**/*',
```
adi-wan-askui, line 62: Do we want to hardcode the versions here? Seems like this would be something that is easily forgotten to be updated with the next releases...

adi-wan-askui (changes requested):
Well done :) Looking forward to the better search :)

PR #584, opened 2023-12-13
docs: enable blog-feature for release notes

JohannesDienst-askui:


adi-wan-askui (changes requested):
Open question from my side: How are release notes tied to the actual releases / versions? As a user, I would want to know with which version / release I actually get these new features.

JohannesDienst-askui:
> Open question from my side: How are release notes tied to the actual releases / versions? As a user, I would want to know with which version / release I actually get these new features.

This is a great question and something @mrmeier-1 and me have to figure out in detail. As there are multiple components with a release cycle of their own.

The current plan is to add one Release Notes entry every week and include everything from the different components under different subheadings with the version.

Does that answer your question @adi-wan-askui ?

adi-wan-askui (approved):

PR #587, opened 2024-01-03
docs: add more explanation to the questions that pop up when you run npm askui init (#DVRL-66)

JohannesDienst-askui:


adi-wan-askui (approved):
Would be awesome if could do something about the duplication so that we only have to change this in one place...

JohannesDienst-askui:
> Would be awesome if could do something about the duplication so that we only have to change this in one place...

This is the biggest pain I have with Markdown -> No includes! With AsciiDoc this would be easy 😉

PR #601, opened 2024-01-22
SSO setup documentation + release notes

mrmeier-1:
- contains a list of identity providers and required information
- link to auth0 docs
- email contact to askui support team

### DIFF
--- a/docs/blog/2024-01-15/2024-01-23.md
+++ b/docs/blog/2024-01-15/2024-01-23.md
@@ -24,7 +24,7 @@ We're introducing two significant enhancements to our platform: Single Sign-On (
 
 #### Single Sign-On Integration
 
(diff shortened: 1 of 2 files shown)


JohannesDienst-askui (changes requested):
Isn't the content there two times. For example the Heading `Setting Up Single Sign-On (SSO) for Your Enterprise Account` with the same paragraph under it is there again in line 72.

Also can we frame this a little bit different. For me WE enable the SSO for the customer and not the other way around. After reading this I got the expression the customer can initiate the SSO setup by themselves. The process is that we enable it with the help of the customer.

JohannesDienst-askui (changes requested):

JohannesDienst-askui (approved):

PR #716, opened 2024-03-04
feat: move credentials to .env file instead of writing them in plain text (#DVRL-71)

JohannesDienst-askui:
We stored the credentials (workspace-id + access token) in `helpers/askui-helpers.ts` in plain text. While this is ok for local projects, it increases the risk of pushing credentials into a repository and exposing them.

This PR moves the credentials into `.env` where they are usually excluded with a proper `.gitignore` minimizing the risk of of leaking credentials 🥳 .

### DIFF
--- a/docs/blog/2024-03-04/2024-03-04.md
+++ b/docs/blog/2024-03-04/2024-03-04.md
@@ -1,0 +1,114 @@
+---
+title: 2024.03.04 | AskUI SDK Convenience Methods
+slug: release-2024-03-04
+authors: [johannesdienst, jonasmenesklou]
+tags: [AskUI SDK]
+hide_table_of_contents: false
+---
+
+## Introduction
+This week we improved on our docs. A guide for setting up your terminal and extensions in Visual Studio Code on Windows will improve your experience while developing AskUI Workflows.
+
+We also fixed a display bug at the _Supported Keys_ page.
+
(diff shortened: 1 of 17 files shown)


mlikasam-askui:
Great job! 💪
So far, we've discussed and agreed not to incorporate dotenv support. Instead, we'll provide documentation on how to implement it as a best practice. This way, users have the flexibility to configure their environment as they see fit.

JohannesDienst-askui:
> Great job! 💪 So far, we've discussed and agreed not to incorporate dotenv support. Instead, we'll provide documentation on how to implement it as a best practice. This way, users have the flexibility to configure their environment as they see fit.

Ah ok. My ticket said I should implement this, or at least I understood it that way 😉 

I would like to install the `dotenv`-dependency anyway and prepare the `.env`-file for the user upon `npm askui init`. The user only has to add the correct import in `askui-helper.ts` for the best practice to work. Is that ok from your side @mlikasam-askui ?

mlikasam-askui:
> > Great job! 💪 So far, we've discussed and agreed not to incorporate dotenv support. Instead, we'll provide documentation on how to implement it as a best practice. This way, users have the flexibility to configure their environment as they see fit.
> 
> Ah ok. My ticket said I should implement this, or at least I understood it that way 😉
> 
> I would like to install the `dotenv`-dependency anyway and prepare the `.env`-file for the user upon `npm askui init`. The user only has to add the correct import in `askui-helper.ts` for the best practice to work. Is that ok from your side @mlikasam-askui ?

Personally, I don't like shipping it like that. However, I would advise waiting for @programminx-askui. You can discuss it with him, and he can then bring it up with @dom-askui. I can't decide for the team.

JohannesDienst-askui:
Moved the docs part to https://github.com/askui/askui-dev-docs/pull/10

PR #718, opened 2024-03-04
Add skip Credentials Flag

mlikasam-askui:
- Include the `--skip-credentials` flag in the `askui init` command.
- Make sure credentials are defined before initiating the execution.

**Jira:** CO-620


JohannesDienst-askui (approved):
Nice! learned a few things here :)

PR #719, opened 2024-03-04
Remove are you proxy user question

mlikasam-askui:
- Remove are you proxy user question from the cli.


JohannesDienst-askui (approved):
Works as intended 🥳

PR #730, opened 2024-03-11
Refactor: Remove npm workspaces config

JohannesDienst-askui:
This is a big change -> We remove the monorepo setup based on `npm` workspaces.

There are two reasons for that. The first is the more important one.

1. We shot ourselves in the foot with it because dependencies leaked into a workspace somehow from `root` causing the askui to fail at runtime but not at build time (Yikes!)
2. We will implement an AskUI SDK in more languages than only TypeScript. Thus the monorepo would not work for them.

What changed:
* Pre-Commit hooks remain in root as they are universally applicable
* `npm` workspaces were removed. Everything that was in `root`  was moved to `packages/askui-nodejs`
* Adapted the README accordingly
* Necessary adaptions in Github-Workflows
* Necessary adaptions in `release-it` config


.husky/pre-commit
```diff
@@ -2,8 +2,9 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx is-ci && exit 0
+cd packages/askui-nodejs
 npm run lint:staged
 npm test
 
-cd examples/askui-nodejs-base-example || exit
+cd ../../examples/askui-nodejs-base-example || exit
```
adi-wan-askui, line 9: Save the root directory into a variable, e.g., `project_root=$(pwd)`, and use it so that you don't have to go up and down the file system tree using relative paths but you use absolute paths instead, e.g., `${project_root}/examples/askui-nodejs-base-example`. This makes these `cd` commands independent of each other and, therefore, omits a lot of potential bugs.

README.md
```diff
@@ -10,15 +10,17 @@
 
 ## Disclaimer
 
-This repo is a [monorepo](https://en.wikipedia.org/wiki/Monorepo#:~:text=In%20version%20control%20systems%2C%20a,as%20a%20'shared%20codebase'.) consisting mainly of npm packages. We use [workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) to managing the different npm packages etc. This may change in the future as we plan to include also packages, libraries etc. written in other languages in this repo to make the power of AskUI available to non-typescript/-javascript developers as well.
+This repo contains the AskUI SDK (ADK) written in TypeScript. Releases are done from the root repository. This may change in the future as we plan to include also packages, libraries etc. written in other languages in this repo to make the power of AskUI available to non-typescript/-javascript developers as well.
 
 ## Installation
-Run an `npm install` inside the root directory.
+Run an `npm install` inside the root directory if you want to create a new release.
```
adi-wan-askui, line 16: I don't think it is super clear for a user what is going on here.

1. I would explain the project structure first which explains why `npm install` is to be run twice at different hierarchy levels. 
2. Then, I would explain what is installed in root (I don't think this  about releasing or at least not only but rather committing etc.) vs. in the `packages/askui-nodejs`.

README.md
```diff
@@ -10,15 +10,17 @@
...
 
 ```sh
 $ npm install
 ```
 
+Run `npm install` inside `packages/askui-nodejs` to build the TypeScript ADK.
```
adi-wan-askui, line 22: Does this already build it? Doesn't it only install the dependencies of it?

README.md
```diff
@@ -31,7 +33,7 @@ Commit messages should conform to [Conventional Commits Message Standard](https:
 
 ### Adding Dependencies
 
-The [one version rule](https://opensource.google/documentation/reference/thirdparty/oneversion#:~:text=There%20may%20only%20be%20one,several%20reasons%20for%20this%20restriction.) should be followed as much as possible. In practice, this mean checking if a dependency to be added is already used by another package or meant to be used by multiple packages. If not, install it inside the corresponding package's root directory, e.g., `./packages/askui-nodejs`. If it is shared, install it in the project's root directory and make sure that you only need to depend on this single version in all packages. 
+The [one version rule](https://opensource.google/documentation/reference/thirdparty/oneversion#:~:text=There%20may%20only%20be%20one,several%20reasons%20for%20this%20restriction.) should be followed as much as possible. In practice, this mean checking if a dependency to be added is already used by another package or meant to be used by multiple packages.
```
adi-wan-askui, line 36: I think we can remove it completely.

examples/askui-nodejs-base-example/.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 dist
 .eslintcache
 report
+package-lock.json
```
adi-wan-askui, line 5: Why? Makes the installation unreproducable...

.husky/pre-commit
JohannesDienst-askui: Sometimes I think too complicated to come up with such an efficient solution 😉

examples/askui-nodejs-base-example/.gitignore
adi-wan-askui: Do you know if we use the example anywhere? Looked to me like this is not the case and we already have a lot of examples in other repos (that also look more up-to-date). So I would just remove the whole `examples` folder.

.husky/pre-commit
```diff
@@ -1,9 +1,11 @@
...
+project_root=$(pwd)
 
 npx is-ci && exit 0
+cd ${project_root}/packages/askui-nodejs
 npm run lint:staged
 npm test
 
-cd examples/askui-nodejs-base-example || exit
+cd ${project_root}/examples/askui-nodejs-base-example || exit
 npm run lint:staged
```
adi-wan-askui, line 11: If you remove `examples`, don't forget to remove this, too, and other places where we reference it.

README.md
```diff
@@ -10,15 +10,30 @@
...
 ## Disclaimer
 
-This repo is a [monorepo](https://en.wikipedia.org/wiki/Monorepo#:~:text=In%20version%20control%20systems%2C%20a,as%20a%20'shared%20codebase'.) consisting mainly of npm packages. We use [workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) to managing the different npm packages etc. This may change in the future as we plan to include also packages, libraries etc. written in other languages in this repo to make the power of AskUI available to non-typescript/-javascript developers as well.
+This repo contains the AskUI SDK (ADK) written in TypeScript. Releases are done from the root repository. This may change in the future as we plan to include also packages, libraries etc. written in other languages in this repo to make the power of AskUI available to non-typescript/-javascript developers as well.
+
+## Repository Structure
+
+At root level we store the configuration for commit hooks, CI/CD and releasing a new version of the ADK.
+
+Under `packages` you find the ADKs for different languages.
```
adi-wan-askui, line 19: Not bad

adi-wan-askui (approved):
I think it is better now :)  Feel free to remove the `examples` in another commit and merge directly or remove them in another PR.

examples/askui-nodejs-base-example/.gitignore
JohannesDienst-askui: Glad that you mentioned it. I second-guessed myself for not knowing what the `examples/askui-nodejs-base-example` is there in the first place 😄 .

I have never seen it used anywhere, so removing it will be unproblematic.

PR #733, opened 2024-03-19
refactor: introduce sort-keys ESLint rule

JohannesDienst-askui:


JohannesDienst-askui:
@programminx-askui The tests for the proxy are failing because the self-signed certificates expired. Are there instructions on how to repair that?

PR #738, opened 2024-04-02
feat: add more convenience methods that are designed with intent in mind

JohannesDienst-askui:
Add/modify convenience methods:

* `clickText({text: string})` -> Breaking change 
* `typeIntoTextfield({textToWrite: string})`
* `clickSwitch({label: string})`
* `clickCheckbox({label: string})`
* `clickButton({})` -> Breaking change

`clickText({})` and `clickButton({})` introduce a braking change as they now take objects instead of a single `string` argument.

Open questions:
* I put `Relations` into `dsl.ts`. It is generated over at `askui-inference`, correct? How do I add this properly to the TypeScript ADK?
* Is the design intuitive enough or is there a better way than using objects as parameters?


packages/askui-nodejs/src/execution/dsl.ts
```diff
@@ -12,6 +12,8 @@ export enum Separators {
   STRING = '<|string|>',
 }
 
+export type Relations = 'nearestTo' | 'leftOf' | 'above' | 'rightOf' | 'below' | 'contains';
```
adi-wan-askui, line 15: You are right, the whole `dsl.ts` file is generated through the askui-inference. @programminx-askui or @mlikasam-askui can help you out with interating it there. Alternatively, you could also simply move this to another file, e.g., `ui-control-client.ts` or a new file.

adi-wan-askui, line 15: I think we should improve on the naming as this does not contain all relations but only a subset that are relevant for a specific use case.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -362,4 +343,300 @@ export class UiControlClient extends ApiCommands {
...
+          break;
+        case 'rightOf':
+          command = command.rightOf().text(params.relation.text);
+          break;
+        case 'below':
+          command = command.below().text(params.relation.text);
+          break;
+        default:
+          command = command.nearestTo().text(params.relation.text);
+          break;
```
adi-wan-askui, line 400: This can only happen if someone is going around the Typescript compiler some way from my perspective using a relation that we don't support. What about throwing an error in this case instead to let the user know so that the user is able to fix this?

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -362,4 +343,300 @@ export class UiControlClient extends ApiCommands {
...
+    },
+  ) {
+    let command = this.click().button();
+
+    if (params.label) {
+      command = command.withText(params.label);
+    }
+
+    if (params.relation) {
+      switch (params.relation.type) {
```
adi-wan-askui, line 382: Minor thing (optional) to improve code readability, maintainability etc.: I would either move switch to a dedicated function or check if `params.relation.type` is included in an array of all the values included in the union type `Relations` and if not, do whatever is under `default` else call `command[params.relation.type]().text(params.relation.text)`. Typescript may not like it but could be disabled for this specific purpose.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -362,4 +343,300 @@ export class UiControlClient extends ApiCommands {
...
+        case 'below':
+          await command.below().text(params.label).exec();
+          break;
+        default:
+          await command.nearestTo().text(params.label).exec();
+          break;
+      }
+    } else {
+      await command.nearestTo().text(params.label).exec();
+    }
```
adi-wan-askui, line 457: Minor thing (optional): I think it would help in the future if these convenience functions were implemented more consistently (with `clickButtton()`,  e.g., calling exec always at the end on the command build previously, maybe even sharing some part of the code, e.g., the relationship building part (⟶ currently, switch statement).

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -362,4 +343,300 @@ export class UiControlClient extends ApiCommands {
...
+   *                                        clicked text and another element.
+   * @property {Relations} params.relation.type - The type of relation.
+   * @property {string} params.relation.text - The label or text associated with the
+   *                                           related element or state.
+   */
+  async clickText(
+    params: {
+      text: string,
+      regex?: boolean,
+      exact?: boolean,
```
adi-wan-askui, line 603: I think that these both correlate with each other can lead to confusing edge cases and increase complexity, e.g., what happens if `regex` and `exact` are both set to `true`. What about combining these two into one string union type, e.g., `type: 'similar' | 'exact' | 'regex' = 'similar'` or `type?: 'exact' | 'regex'`?

adi-wan-askui (changes requested):
I like the switch from simple string params to objects as this gives more flexibility in the future and simulates keyword arguments of other languages, e.g., Python, that a lot of our users highly enjoy. It is unfortunate that this is a breaking change but I think it is fine as it is still quite new and not many people have probably adapted it thoroughly.

packages/askui-nodejs/src/execution/ui-control-client.ts
JohannesDienst-askui: Good idea. Implemented it this way now :)

packages/askui-nodejs/src/execution/dsl.ts
JohannesDienst-askui: Not sure if `Anchor` is a better name for it, but we used it already in the docs and it makes kinda sense.

I am open to better suggestions.

JohannesDienst-askui:
@adi-wan-askui I hope all comments got addressed 😃 

Thanks for the great review!

packages/askui-nodejs/src/execution/dsl.ts
adi-wan-askui: Nice term but I think it may be ambiguous and like you said, we already use the term `Relations` extensively. As this is a subset of all possible relations, what about a name like `RelationsForConvenienceMethods` (also clunky, maybe you can think of a better prefix but the general idea is to use `Relations...` prefix and specify the subset afterwards)?

packages/askui-nodejs/src/execution/ui-control-client.ts
adi-wan-askui: Very nice approach :)

packages/askui-nodejs/src/execution/dsl.ts
JohannesDienst-askui: My first programming language was Java, so clunkiness is not a problem for me 😉 .

I like your thinking there and would go with your suggestion `RelationsForConvenienceMethods`. Better clunky and expressive than not clunky and ambiguous.

JohannesDienst-askui:
@adi-wan-askui Pushed the final edits. Waiting for your approval 😃 .

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -344,6 +348,38 @@ export class UiControlClient extends ApiCommands {
...
+        commando = command.below().text(text);
+        break;
+      case 'contains':
+        commando = command.contains().text(text);
+        break;
+      default:
+        throw new ValidationError('No valid Relation.Type was passed.');
+    }
+    return commando;
+  }
```
adi-wan-askui, line 381: Just some minor ideas for improving the expressiveness and conciseness of the code (optional):

```suggestion
  private extendCommandWithRelationToText(
    command: FluentFiltersOrRelations,
    relation: RelationsForConvenienceMethods,
    text: string,
  ) {
    switch (relation) {
      case 'nearestTo':
        return command.nearestTo().text(text);
      case 'leftOf':
        return command.leftOf().text(text);
      case 'above':
        return command.above().text(text);
      case 'rightOf':
        return command.rightOf().text(text);
      case 'below':
        return command.below().text(text);
      case 'contains':
        return command.contains().text(text);
      default:
        throw new ValidationError(`Relation "{relation}" unknown. Please, use one of the following relations: {relationsForConvenienceMethods.map(r => `"{r}"`).join(", ")}`).;
    }
  }
```

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -574,67 +555,57 @@ export class UiControlClient extends ApiCommands {
...
+        command = command.withText(params.text);
+        break;
+      case 'exact':
+        command = command.withExactText(params.text);
+        break;
+      case 'regex':
+        command = command.withTextRegex(params.text);
+        break;
+      default:
+        throw new ValidationError('"type" must be "similar", "exact" or "regex"');
```
adi-wan-askui, line 604:
```suggestion
        throw new ValidationError(`Type "{params.type}" unknown. Please, use one of the following types: "similar", "exact", "regex".`);
```

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -574,67 +555,57 @@ export class UiControlClient extends ApiCommands {
...
    * @property {string} params.relation.text - The label or text associated with the
    *                                           related element or state.
    */
   async clickText(
     params: {
       text: string,
-      regex?: boolean,
-      exact?: boolean,
-      relation?: { type: Relations, text: string }
+      type: 'similar' | 'exact' | 'regex',
```
adi-wan-askui, line 584: What about making it optional or moving it below `relation` so you can add `'similar'` as the default?

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -574,67 +555,57 @@ export class UiControlClient extends ApiCommands {
...
    *                          for regular expression matching and relation.
    * @property {string} params.text - The text to be clicked.
-   * @property {boolean} [params.regex=false] - Whether the text is matched using a regular
-   *                                            expression, default is false.
-   * @property {boolean} [params.exact=false] - Whether an exact match of the text is required
-   *                                            when using regex, default is false.
-   * @property {Object} [params.relation] - Object describing the relationship between the
-   *                                        clicked text and another element.
-   * @property {Relations} params.relation.type - The type of relation.
+   * @property {string} params.type - Whether the text is matched using similarity,
```
adi-wan-askui, line 573:
```suggestion
   * @property {('similar' | 'exact' | 'regex')} params.type - Whether the text is matched using similarity,
```

adi-wan-askui (approved):
Nicely done :) Just found some minor optimisation potentials. Feel free to skip.

PR #740, opened 2024-04-10
chore: track if running in ci (#CL-682)

adi-wan-askui:


programminx-askui:
@JohannesDienst-askui 

```
How do we collect your data?

Your data is collected, on the one hand, by the fact that you provide it to us. This can be, for example, data that you enter in a contact form.

Other data is collected automatically or after your consent when visiting the website by our IT systems. These are mainly technical data (e. e. Internet browser, operating system or time of page view). The collection of this data takes place automatically as soon as you enter this website.
What do we use your data for?

Some of the data is collected to ensure error-free provision of the website. Other data may be used to analyze your user behavior.
```
https://www.askui.com/data-privacy

And the Consent is get after Signing in. 

There could be some improvements to extend the Data Privacy, but should be fine for now.

PR #744, opened 2024-04-18
feat(ML-492): use V4 online learning endpoint.

danyalxahid-askui:
# Description

Supports the v4 version of online learning of inference. Backward compatible.

# Tested on
- Local test with only using adding "inferenceServerApiVersion" as "v4" inside UiControlClient.build({})
- Used Dart application to train the model
- Call the test case again and annotate in html if the difference is there

### DIFF
--- a/packages/askui-nodejs/src/execution/inference-client.ts
+++ b/packages/askui-nodejs/src/execution/inference-client.ts
@@ -27,7 +27,7 @@ export class InferenceClient {
     private readonly resize?: number,
     readonly workspaceId?: string,
     readonly modelComposition?: ModelCompositionBranch[],
-    private readonly apiVersion = 'v4',
+    private readonly apiVersion = 'v3',
   ) {
     const versionedBaseUrl = urljoin(this.baseUrl, 'api', this.apiVersion);
     const url = workspaceId
--- a/packages/askui-nodejs/src/execution/inference-client.ts
+++ b/packages/askui-nodejs/src/execution/inference-client.ts
@@ -48,11 +48,6 @@ export class InferenceClient {
   }
 
   async isImageRequired(instruction: string): Promise<boolean> {
-    // V4's image is always required
-    if (this.urls.inference.endsWith('predict')) {
-      return true;
-    }
-
     const response = await this.httpClient.post<IsImageRequired>(
       this.urls.isImageRequired,
       {
--- a/packages/askui-nodejs/src/execution/inference-client.ts
+++ b/packages/askui-nodejs/src/execution/inference-client.ts
@@ -48,9 +48,9 @@ export class InferenceClient {
   }
 
   async isImageRequired(instruction: string): Promise<boolean> {
-    // V4's image is not required
+    // V4's image is always required
     if (this.urls.inference.endsWith('predict')) {
-      return false;
+      return true;
     }
 
     const response = await this.httpClient.post<IsImageRequired>(
@@ -83,7 +83,7 @@ export class InferenceClient {
       this.urls.inference,
       this.urls.inference.endsWith('predict') ? {
         // V4 only supports the base64 image and not the metadata of the base64 string
-        image: resizedImage.base64Image?.split(',')[1],
+        image: resizedImage.base64Image,
         instruction,
         tasks: ['OCR'],
       }
--- a/packages/askui-nodejs/src/execution/inference-client.ts
+++ b/packages/askui-nodejs/src/execution/inference-client.ts
@@ -27,7 +27,7 @@ export class InferenceClient {
     private readonly resize?: number,
     readonly workspaceId?: string,
     readonly modelComposition?: ModelCompositionBranch[],
-    private readonly apiVersion = 'v3',
+    private readonly apiVersion = 'v4',
   ) {
     const versionedBaseUrl = urljoin(this.baseUrl, 'api', this.apiVersion);
     const url = workspaceId
--- a/packages/askui-nodejs/src/execution/inference-client.ts
+++ b/packages/askui-nodejs/src/execution/inference-client.ts
@@ -35,7 +35,7 @@ export class InferenceClient {
       : versionedBaseUrl;
     this.urls = {
       // V4's only end part is different
-      inference: apiVersion === 'v4'? urljoin(url, 'predict') : urljoin(url, 'inference'),
+      inference: apiVersion === 'v4' ? urljoin(url, 'predict') : urljoin(url, 'inference'),
       isImageRequired: urljoin(url, 'instruction', 'is-image-required'),
     };
     this.httpClient.urlsToRetry = Object.values(this.urls);
@@ -49,9 +49,9 @@ export class InferenceClient {
 
   async isImageRequired(instruction: string): Promise<boolean> {
     // V4's image is not required
-    if(this.urls.inference.endsWith("predict")){
+    if (this.urls.inference.endsWith('predict')) {
       return false;
-    }    
+    }
 
     const response = await this.httpClient.post<IsImageRequired>(
       this.urls.isImageRequired,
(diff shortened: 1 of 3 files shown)


packages/askui-nodejs/src/execution/inference-client.ts
```diff
@@ -75,12 +81,18 @@ export class InferenceClient {
     const resizedImage = await this.resizeIfNeeded(customElements, image);
     const response = await this.httpClient.post<InferenceResponseBody>(
       this.urls.inference,
-      {
-        customElements,
-        image: resizedImage.base64Image,
+      this.urls.inference.endsWith('predict') ? {
+        // V4 only supports the base64 image and not the metadata of the base64 string
+        image: resizedImage.base64Image?.split(',')[1],
```
mlikasam-askui, line 86: Should we move this task to the backend, like we do [here](https://github.com/askui/askui-inference/blob/dbe7d649e72a2681cf19dd066245ee3f1a31ed1d/controlyourui/helper.py#L24)?

packages/askui-nodejs/src/execution/inference-client.ts
```diff
@@ -34,7 +34,8 @@ export class InferenceClient {
       ? urljoin(versionedBaseUrl, 'workspaces', workspaceId)
       : versionedBaseUrl;
     this.urls = {
-      inference: urljoin(url, 'inference'),
+      // V4's only end part is different
+      inference: apiVersion === 'v4' ? urljoin(url, 'predict') : urljoin(url, 'inference'),
```
mlikasam-askui, line 38: why are you not using 'inference' anymore?

packages/askui-nodejs/src/execution/inference-client.ts
```diff
@@ -47,6 +48,11 @@ export class InferenceClient {
   }
 
   async isImageRequired(instruction: string): Promise<boolean> {
+    // V4's image is not required
+    if (this.urls.inference.endsWith('predict')) {
+      return false;
+    }
```
mlikasam-askui, line 54: Could you clarify why an image isn't necessary? This means that no screenshot will be captured, and the inference request won't include any image.

packages/askui-nodejs/src/execution/inference-client.ts
```diff
@@ -75,12 +81,18 @@ export class InferenceClient {
...
-      {
-        customElements,
-        image: resizedImage.base64Image,
+      this.urls.inference.endsWith('predict') ? {
+        // V4 only supports the base64 image and not the metadata of the base64 string
+        image: resizedImage.base64Image?.split(',')[1],
         instruction,
-        modelComposition: this.modelComposition,
-      },
+        tasks: ['OCR'],
```
mlikasam-askui, line 88: This will lead to only detecting text elements? Is this the intended behavior?

packages/askui-nodejs/src/execution/inference-client.ts
danyalxahid-askui: Fixed. Image is always required for the "v4" implementation.

packages/askui-nodejs/src/execution/inference-client.ts
danyalxahid-askui: Yes.

We moved away from Triton based model inference to a container based deploytment which serves only 1 TASK (initially started off as OCR but backend is build in a way which is easily extendable).

You can find the details in this [EPIC](https://askyourui.atlassian.net/browse/ML-463) and its related Stories.

You can find other implementations:
- Backend Implementation in this [PR](https://github.com/askui/askui-inference/pull/186).
- Infrastructure in this [folder](https://github.com/askui/infrastructure/tree/main/clusters/askui-dev-v2/default/end-to-end-ocr)
- Desktop Application for Flutter in this [Repository](https://github.com/askui/ml-ocr-text-recognition-desktop-frontend)


Also the working solution excluding the Askui-lib can be found [here](https://www.notion.so/askui/007_Dev_ML_Update-a8236e2a690f4e9f9bf202eb1f57871a?pvs=4#b2f06ddfa5204c65bdc4ac5e8cc57e43)

packages/askui-nodejs/src/execution/inference-client.ts
danyalxahid-askui: We decide to move away from the current implementation of model composition and URL to make it simpler and user readable. Can be found in [Ticket](https://askyourui.atlassian.net/browse/ML-465)

packages/askui-nodejs/src/execution/inference-client.ts
danyalxahid-askui: You the Inference [PR](https://github.com/askui/askui-inference/pull/186) if you want.

The only thing remaining is to move Threading to ThreadPool executor which should be completed today.

danyalxahid-askui: @mlikasam-askui Move threading to Threadpool executor as done in the main branch in "detectors_event_pipes.py". Deployed changes to Infrastructure (without merging).

mlikasam-askui: Please correct me if I'm mistaken, but from what I understand:

Currently, users are encountering issues with OCR accuracy. To address this, you've implemented a solution to correct OCR inaccuracies, with the intention of allowing users to utilize the newly trained OCR.
However, there's a current limitation where users are restricted to only using text elements because the inference graph doesn't include other models. Consequently, any previous tests involving elements other than text, such as clicking on buttons, will fail.
Additionally, there's a constraint where only one OCR version is allowed per workspace ID.
 
For the "is-image-required" endpoint, if you're just moving the mouse to coordinates like (x=2, y=2), there's no need for an image or OCR. That's why the endpoint exists—to skip unnecessary image taking and sending.

What do you think about keeping the current inference graph setup, but swapping out the OCR with a trained one if it exsits? This way, we can minimize changes in the ADK since it still operates under the same logic.

danyalxahid-askui: Yes, You understood most things correctly.

1.) Thanks for the explanation on the "is-image-required" endpoint. Fixed it.

2.) Any previous test will continue to work as they did. The current solution is intended for backward compatibility. This is because:
- There are 2 separate API versions. V3 consists of everything that was done previously and no changes have been implemented.
- The default version used for every initialization of the test (including previous tests written) will point to "V3" implementation.
- Only the new implementation where we specify the "inferenceServerApiVersion" as "v4" will go through the new process.

3.) Yes, The constraint for 1 trained model per workspace is intended. Every workspace will have its separate trained model. We are not training the base model for all the users right now. In later stages, we can use the data provided by the users to train the main model and update the base model if necessary.

4.) We initially did release a version to [prod](https://ml-ocr.askui.com/) where we had a streamlit application doing the same and the new model was being served from triton which was workspace specific. There is a limitation of triton that when we update the model e.g. the draft model it still did not update in inference because triton caches the initially loaded model from S3. And if we update the draft model it does not get reflected until we restart triton. This is why @MichaelMayer-askui added restarting the triton server every night [here](https://github.com/askui/infrastructure/blob/main/clusters/askui-prod-v2/default/triton-inference-server-gpu-cron.yaml) which is not optimal.
- The process of why we do things this way is because we want to get rid of triton and move everything to a containerized-based approach in which 1 container is responsible for 1 Task (OD, OCR...). Can be found in [This Epic](https://askyourui.atlassian.net/browse/ML-439) followed by [This Epic](https://askyourui.atlassian.net/browse/ML-463) in detail.
- This is just the first step in moving all models to the containerized-based approach.
- In the next steps we also want to add a new Decision Logic other than UIDT1. On Runtime the User will be able to use Different Decision logic to what feels best for the user's needs e.g. [PTA-TEXT](https://askyourui.atlassian.net/browse/ML-467)


Let me know if you want to schedule a call and make things more clearer.

packages/askui-nodejs/src/execution/inference-client.ts
```diff
@@ -34,7 +34,7 @@ export class InferenceClient {
       ? urljoin(versionedBaseUrl, 'workspaces', workspaceId)
       : versionedBaseUrl;
     this.urls = {
-      inference: urljoin(url, 'inference'),
+      inference: apiVersion === 'v4-experimental' ? urljoin(url, 'predict') : urljoin(url, 'inference'),
```
programminx-askui, line 37:
```suggestion
```

programminx-askui (changes requested):

programminx-askui (approved):

packages/askui-nodejs/src/execution/inference-client.ts
danyalxahid-askui: Fixed

PR #746, opened 2024-04-24
feat: change eslint-rule-askui to @askui/eslint-rule-askui; Activate the rule for expect

JohannesDienst-askui:
* Activate the new ESLint rule for correct `expect()` instructions
* Noticed we still had the old dependency reference (not in @askui org) when we `init` askui. Changed it to the org one.

### DIFF
--- a/packages/askui-nodejs/example_projects_templates/typescript/.eslintrc.json-template
+++ b/packages/askui-nodejs/example_projects_templates/typescript/.eslintrc.json-template
@@ -15,7 +15,7 @@
     "@askui/askui"
   ],
   "rules": {
-      "@askui/askui/no-missing-askui-exec": "error",
-      "@askui/askui/correct-askui-expect-usage": "error"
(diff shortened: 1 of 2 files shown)


packages/askui-nodejs/example_projects_templates/typescript/.eslintrc.json-template
```diff
@@ -12,9 +12,10 @@
   },
   "plugins": [
     "import",
-    "askui"
+    "@askui/askui"
   ],
   "rules": {
-      "askui/no-missing-askui-exec": "error"
+      "@askui/askui/no-missing-askui-exec": "error",
+      "@askui/askui/correct-askui-expect-usage": "error"
```
adi-wan-askui, line 19: Can we simplify the names? E.g.,

```suggestion
      "@askui/askui/no-missing-exec": "error",
      "@askui/askui/expect": "error"
```

adi-wan-askui (approved):
Good job :)

PR #750, opened 2024-05-31
CL-728 Track askui client id and client session id

adi-wan-askui:
- Client id: Id of instance of the `UiControlClient` created at runtime
- Client session id: Id of session (e.g., started by `npm run test`) in which client is used

Limitation: Session id currently different for each process if user decides to parallelise with worker processes

### DIFF
--- a/packages/askui-nodejs/src/utils/analytics/analytics.ts
+++ b/packages/askui-nodejs/src/utils/analytics/analytics.ts
@@ -1,4 +1,5 @@
 import os from 'os';
+import crypto from 'crypto';
 import { AnalyticsInterface } from './analytics-interface';
 import { UserIdentifierInterface } from './user-identifier-interface';
 import { UserIdentifier } from './user-identifier';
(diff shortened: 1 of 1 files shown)


packages/askui-nodejs/src/utils/analytics/analytics.ts
```diff
@@ -1,16 +1,23 @@
...
+  private static clientSessionId: string = crypto.randomUUID();
+
+  private clientId: string = crypto.randomUUID();
+
   private userIdentifier: UserIdentifierInterface = new UserIdentifier();
 
   async getAnalyticsHeaders(context: Context): Promise<Record<string, string>> {
     const userID = await this.userIdentifier.userId();
     const headers: Record<string, string> = {
+      'askui-client-id': this.clientId,
```
danyalxahid-askui, line 19: Just to understand.

We are using "clientSessionId" which is same throughout the session across all test cases and "clientId" which will be different for each test case?

If this is the case why do we need "clientId".


If "clientSessionId" is used across the whole session and with several runs what does the purpose of clientID fulfill? Is it to analyze if all test cases were passed in a session after test cases were fixed?

adi-wan-askui: A _session_ is defined a one test run, i.e., the user executing `npm run ...`, which may contain 1..* test cases. The _client session id_ is the same across the whole session for all instances of the `UiControlClient`, often called `aui`, i.e., it identifies the session. The reason why it is not simply called  _session id_ is that it is bound to the client in a scenario where the user would use multiple worker processes it would be different for each worker process.  The _client id_ on the other hand identifies the instance of the `UiControlClient`. It is possible to use multiple instances of the `UiControlClient` within the same session, e.g., to control different devices at the same time. In this case, the client id would be different for each instance but the client session id would be the same for all. Consequently, we can identify both, the session (except for when the user uses worker processes which is one of the current limitations) and the client, e.g., if the user is using multiple clients.

danyalxahid-askui (approved):

PR #751, opened 2024-06-06
feat: convenience method expectAllExist() for checking if multiple elements are detected (#APD-483)

JohannesDienst-askui:
Adds the convenience method

```async expectExistence(
    params: ExpectExistenceInputParameter[],
  ): Promise<ExpectExistenceReturnValue>
  ```
  
  It takes an array of elements in the form of `ExpectExistenceInputParameter`:
  
  ```
  export type ExpectExistenceInputParameter = {
  type: ExpectExistenceAllowedElementClasses;
  label: string;
  matching?: 'similar' | 'exact' | 'regex';
  relation?: {
    type: RelationsForConvenienceMethods;
    text?: string;
  };
  exists?: boolean;
};
```

and returns `ExpectExistenceReturnValue` with a property `everythingExists` and the the Input-Parameter with `exists` that shows if the element was found:

```
export type ExpectExistenceReturnValue = {
  everythingExists: boolean;
  elements: ExpectExistenceInputParameter[];
};
```

Comprehensive tests can be found in our QA-Repository: https://github.com/askui/askui-qa-system-test/pull/2

### DIFF
--- a/packages/askui-nodejs/src/execution/ui-control-client.ts
+++ b/packages/askui-nodejs/src/execution/ui-control-client.ts
@@ -21,7 +21,7 @@ import { Instruction, StepReporter } from '../core/reporting';
 export type RelationsForConvenienceMethods = 'nearestTo' | 'leftOf' | 'above' | 'rightOf' | 'below' | 'contains';
 
 export type ExpectExistenceAllowedElementClasses = 'text' | 'button' | 'checkbox' | 'switch' | 'textfield' | 'element';
-export type ExpectExistenceInputParameter = {
+export interface ExpectExistenceInputParameter {
   type: ExpectExistenceAllowedElementClasses;
   label: string;
   matching?: 'similar' | 'exact' | 'regex';
@@ -29,11 +29,13 @@ export type ExpectExistenceInputParameter = {
     type: RelationsForConvenienceMethods;
     text?: string;
   };
-  exists?: boolean;
 };
+export interface ExpectExistenceElement extends ExpectExistenceInputParameter {
+  exists: boolean;
+}
 export type ExpectExistenceReturnValue = {
   everythingExists: boolean;
-  elements: ExpectExistenceInputParameter[];
+  elements: ExpectExistenceElement[];
 };
 
 export class UiControlClient extends ApiCommands {
@@ -624,16 +626,11 @@ export class UiControlClient extends ApiCommands {
   // eslint-disable-next-line class-methods-use-this
   private async evaluateFinalExpectExistenceCommand(
     finalCommand: FluentFiltersOrRelationsGetter,
-    element: ExpectExistenceInputParameter,
-  ): Promise<ExpectExistenceInputParameter> {
-    const el = element;
-    el.exists = false;
+  ): Promise<boolean> {
     if ((await finalCommand.exec()).length > 0) {
-      el.exists = true;
+      return true;
     }
-    return new Promise<ExpectExistenceInputParameter>((resolve) => {
-      resolve(el);
-    });
+    return false;
   }
 
   // eslint-disable-next-line class-methods-use-this
@@ -767,7 +764,7 @@ export class UiControlClient extends ApiCommands {
           }
 
           returnValue.elements.push(
-            (await this.evaluateFinalExpectExistenceCommand(finalCommand, element)),
+            {...element, exists: await this.evaluateFinalExpectExistenceCommand(finalCommand)},
           );
           break;
         }
@@ -788,7 +785,7 @@ export class UiControlClient extends ApiCommands {
           }
 
           returnValue.elements.push(
-            (await this.evaluateFinalExpectExistenceCommand(finalCommand, element)),
+            {...element, exists: await this.evaluateFinalExpectExistenceCommand(finalCommand)},
           );
           break;
         }
@@ -807,7 +804,7 @@ export class UiControlClient extends ApiCommands {
           }
 
           returnValue.elements.push(
-            (await this.evaluateFinalExpectExistenceCommand(finalCommand, element)),
+            {...element, exists: await this.evaluateFinalExpectExistenceCommand(finalCommand)},
           );
           break;
         }
@@ -834,7 +831,7 @@ export class UiControlClient extends ApiCommands {
           }
 
           returnValue.elements.push(
-            (await this.evaluateFinalExpectExistenceCommand(finalCommand, element)),
+            {...element, exists: await this.evaluateFinalExpectExistenceCommand(finalCommand)},
           );
           break;
         }
--- a/packages/askui-nodejs/src/execution/ui-control-client.ts
+++ b/packages/askui-nodejs/src/execution/ui-control-client.ts
@@ -4,7 +4,6 @@ import {
   Exec,
   Executable,
   FluentFilters,
-  FluentFiltersGetter,
   ApiCommands,
   Separators,
   PC_AND_MODIFIER_KEY,
@@ -621,11 +620,19 @@ export class UiControlClient extends ApiCommands {
     await command.exec();
   }
 
+  private evaluateMatchingProperty(
+    command: FluentFiltersOrRelations,
+    text: ExpectExistenceInputParameterText,
+  ): FluentFiltersOrRelations;
+  private evaluateMatchingProperty(
+    command: FluentFiltersOrRelationsGetter,
+    text: ExpectExistenceInputParameterText,
+  ): FluentFiltersOrRelationsGetter;
   // eslint-disable-next-line class-methods-use-this
   private evaluateMatchingProperty(
-    command: FluentFiltersGetter | FluentFiltersOrRelationsGetter | FluentFiltersOrRelations,
+    command: FluentFiltersOrRelations | FluentFiltersOrRelationsGetter,
     text: ExpectExistenceInputParameterText,
-  ): FluentFiltersOrRelationsGetter | FluentFiltersOrRelations {
+  ): FluentFiltersOrRelations | FluentFiltersOrRelationsGetter {
     switch (text.matching ?? 'similar') {
       case 'exact':
         return command.withExactText(text.value);
@@ -702,7 +709,7 @@ export class UiControlClient extends ApiCommands {
       const acc = await accumulatorPromise;
       const command = this.get()[param.type]();
       let finalCommand = param.text !== undefined
-        ? this.evaluateMatchingProperty(command, param.text) as FluentFiltersOrRelationsGetter
+        ? this.evaluateMatchingProperty(command, param.text)
         : command;
       if (param.relation) {
         finalCommand = this.evaluateRelation(
--- a/packages/askui-nodejs/src/execution/ui-control-client.ts
+++ b/packages/askui-nodejs/src/execution/ui-control-client.ts
@@ -1,8 +1,12 @@
-import ValidationError from 'yup/lib/ValidationError';
 import { CustomElement } from '../core/model/custom-element';
 import { CustomElementJson } from '../core/model/custom-element-json';
 import {
-  Exec, Executable, FluentFilters, ApiCommands, Separators,
+  Exec,
+  Executable,
+  FluentFilters,
+  FluentFiltersGetter,
+  ApiCommands,
+  Separators,
   PC_AND_MODIFIER_KEY,
   FluentFiltersOrRelations,
   FluentFiltersOrRelationsGetter,
@@ -19,17 +23,21 @@ import { UiControlClientDependencyBuilder } from './ui-control-client-dependency
 import { Instruction, StepReporter } from '../core/reporting';
 
 export type RelationsForConvenienceMethods = 'nearestTo' | 'leftOf' | 'above' | 'rightOf' | 'below' | 'contains';
-
-export type ExpectExistenceAllowedElementClasses = 'text' | 'button' | 'checkbox' | 'switch' | 'textfield' | 'element';
+export type TextMatchingOption = 'similar' | 'exact' | 'regex';
+export type ExpectExistenceInputParameterType = 'otherElement' | 'switch' | 'element' | 'container' | 'checkbox' | 'element' | 'button' | 'table' | 'text' | 'icon' | 'image' | 'textfield';
+export interface ExpectExistenceInputParameterText {
+  value: string;
+  matching?: TextMatchingOption;
+}
+export interface ExpectExistenceInputParameterRelation {
+  type: RelationsForConvenienceMethods;
+  text: string;
+}
 export interface ExpectExistenceInputParameter {
-  type: ExpectExistenceAllowedElementClasses;
-  label: string;
-  matching?: 'similar' | 'exact' | 'regex';
-  relation?: {
-    type: RelationsForConvenienceMethods;
-    text?: string;
-  };
-};
+  type: keyof Pick<FluentFilters, ExpectExistenceInputParameterType>;
+  text?: ExpectExistenceInputParameterText;
+  relation?: ExpectExistenceInputParameterRelation;
+}
 export interface ExpectExistenceElement extends ExpectExistenceInputParameter {
   exists: boolean;
 }
@@ -372,31 +380,21 @@ export class UiControlClient extends ApiCommands {
     command: FluentFiltersOrRelations | FluentFiltersOrRelationsGetter,
     relation: RelationsForConvenienceMethods,
     text: string,
-  ) {
-    let commando = command;
+  ): FluentFiltersOrRelations | FluentFiltersOrRelationsGetter {
     switch (relation) {
-      case 'nearestTo':
-        commando = command.nearestTo().text(text);
-        break;
       case 'leftOf':
-        commando = command.leftOf().text(text);
-        break;
+        return command.leftOf().text(text);
       case 'above':
-        commando = command.above().text(text);
-        break;
+        return command.above().text(text);
       case 'rightOf':
-        commando = command.rightOf().text(text);
-        break;
+        return command.rightOf().text(text);
       case 'below':
-        commando = command.below().text(text);
-        break;
+        return command.below().text(text);
       case 'contains':
-        commando = command.contains().text(text);
-        break;
+        return command.contains().text(text);
       default:
-        throw new ValidationError('No valid Relation.Type was passed.');
+        return command.nearestTo().text(text);
     }
-    return commando;
   }
 
   /**
@@ -594,7 +592,7 @@ export class UiControlClient extends ApiCommands {
    * @param {Object} params - Object containing required `text` property and optional properties
    *                          for regular expression matching and relation.
    * @property {string} params.text - The text to be clicked.
-   * @property {string} params.type - Whether the text is matched using similarity,
+   * @property {string} params.matching - Whether the text is matched using similarity,
    *                                  exact match or a regular expression.
    * @property {Object} params.relation - Object describing the relationship between the
    *                                      clicked text and another element.
@@ -605,7 +603,7 @@ export class UiControlClient extends ApiCommands {
   async clickText(
     params: {
       text: string,
-      type: 'similar' | 'exact' | 'regex',
+      matching: TextMatchingOption,
       relation?: {
         type: RelationsForConvenienceMethods,
         text: string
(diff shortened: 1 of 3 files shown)


packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -19,6 +20,22 @@ import { Instruction, StepReporter } from '../core/reporting';
 
 export type RelationsForConvenienceMethods = 'nearestTo' | 'leftOf' | 'above' | 'rightOf' | 'below' | 'contains';
 
+export type ExpectExistenceAllowedElementClasses = 'text' | 'button' | 'checkbox' | 'switch' | 'textfield' | 'element';
+export type ExpectExistenceInputParameter = {
+  type: ExpectExistenceAllowedElementClasses;
+  label: string;
```
adi-wan-askui, line 26: Why is the `label` not optional? To me it seems like the `type` is enough.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -549,7 +570,7 @@ export class UiControlClient extends ApiCommands {
 
   /**
    * Click on a specific text.
-   * You can also use a RegEx or match the text exactly by specifyicing the specific flag.
+   * You can also use a RegEx or match the text exactly by specifying the specific flag.
```
adi-wan-askui, line 573: Nice catches :)

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
+   * @property {RelationsForConvenienceMethods} params.relation.type - The type of relation.
+   * @property {string} params.relation.text - The label or text associated with the
+   *                                           related element or state.
+   * @returns {ExpectExistenceReturnValue} - Property boolean 'everythingExists' tells
+   *                                         if every element exists.
+   */
+  async expectExistence(
+    params: ExpectExistenceInputParameter[],
+  ): Promise<ExpectExistenceReturnValue> {
+    if (params === undefined) {
```
adi-wan-askui, line 743: The error is not matching the condition. You are testing if `params` is `undefined` but you according to the error message, you seem like you want to be testing if params is an empty array. 

For the test if this is undefined, you could rely on typescript (cannot be undefined). If you want to test it as the user may not use typescript, I would have two conditions with a different validation error message for each.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
+    }
+    return new Promise<ExpectExistenceInputParameter>((resolve) => {
+      resolve(el);
+    });
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  private async evaluateMatchingProperty(
+    command: FluentFiltersOrRelationsGetter | FluentFiltersOrRelations,
+    matching: 'similar' | 'exact' | 'regex' | undefined,
```
adi-wan-askui, line 642: It does not look like this can be `undefined`.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
-        command = command.withExactText(params.text);
+        commando = command.withExactText(text);
         break;
       case 'regex':
-        command = command.withTextRegex(params.text);
+        commando = command.withTextRegex(text);
         break;
       default:
-        throw new ValidationError('"type" must be "similar", "exact" or "regex"');
+        throw new ValidationError('"matching" property must be "similar", "exact" or "regex"');
```
adi-wan-askui, line 657: Why not just have a default value. To me it looked like this should be impossible. I would test early at the beginning of the public API if you want to be defensive.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
+    if ((await finalCommand.exec()).length > 0) {
+      el.exists = true;
+    }
+    return new Promise<ExpectExistenceInputParameter>((resolve) => {
+      resolve(el);
+    });
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  private async evaluateMatchingProperty(
```
adi-wan-askui, line 640: I think this does not need to be an async function. At least I don't see why. You can directly return from the cases of the switch statement. That would make the method way less complex.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
+      const element = params[i] as ExpectExistenceInputParameter;
+
+      switch (element.type) {
+        case 'text': {
+          let command = this.get().text();
+          command = await this.evaluateMatchingProperty(
+            command,
+            element.matching,
+            element.label,
+          ) as FluentFiltersOrRelationsGetter;
```
adi-wan-askui, line 762: If you really need a type cast, I would put it into the method and use the correct return type. But ideally, I would try to omit it by fixing the typing.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
-      command = this.evaluateRelation(command, params.relation.type, params.relation.text);
+  private evaluateRelationExpectExistence(
+    element: ExpectExistenceInputParameter,
+    command: FluentFiltersOrRelationsGetter,
+  ) {
+    if (element.relation === undefined) {
+      throw new ValidationError(
+        '"relation" must never be undefined when searching for a switch or checkbox.',
+      );
     }
```
adi-wan-askui, line 672: How do you know here if it is a `switch` statement or `checkbox`? This is the wrong place for this validation from my perspective...

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
+    // param.relation.text must never be undefined here -> User error
```
adi-wan-askui, line 673: Wrong place for this validation... This method knows more than it should know...

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
+    if (element.relation.text === undefined) {
+      throw new ValidationError(
+        '"relation.text" must never be undefined when searching for a switch or checkbox.',
+      );
+    }
+    return this.evaluateRelation(
+      command,
+      element.relation.type,
+      element.relation.text,
+    ) as FluentFiltersOrRelationsGetter;
```
adi-wan-askui, line 683: Why do you need a type cast?

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
+      resolve(commando);
+    });
+  }
 
-    if (params.relation) {
-      command = this.evaluateRelation(command, params.relation.type, params.relation.text);
+  private evaluateRelationExpectExistence(
+    element: ExpectExistenceInputParameter,
+    command: FluentFiltersOrRelationsGetter,
+  ) {
```
adi-wan-askui, line 667: Return type?

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -588,26 +610,247 @@ export class UiControlClient extends ApiCommands {
...
+   *                                      exact match or a regular expression.
+   * @property {Object} params.relation - Object describing the relationship between the
+   *                                      clicked text and another element.
+   * @property {RelationsForConvenienceMethods} params.relation.type - The type of relation.
+   * @property {string} params.relation.text - The label or text associated with the
+   *                                           related element or state.
+   * @returns {ExpectExistenceReturnValue} - Property boolean 'everythingExists' tells
+   *                                         if every element exists.
+   */
+  async expectExistence(
```
adi-wan-askui, line 740: This method is way too long and (unnecessarily) complex from my perspective starting with its signature aka the new addition to the api... I think we should have a quick call. Maybe I am simply lacking context...

adi-wan-askui (changes requested):
First of all some likes:
- Test suites
- In code documentation very thorough

But I think the api and code is really complex for what it tries to achieve... Not sure if I am to review the api here, too, though. But I think we should have a quick call in order not to go back and forth because I think I may be lacking context.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -19,21 +23,27 @@ import { UiControlClientDependencyBuilder } from './ui-control-client-dependency
...
+export type ExpectExistenceInputParameterType = 'otherElement' | 'switch' | 'element' | 'container' | 'checkbox' | 'element' | 'button' | 'table' | 'text' | 'icon' | 'image' | 'textfield';
+export interface ExpectExistenceInputParameterText {
+  value: string;
+  matching?: TextMatchingOption;
+}
+export interface ExpectExistenceInputParameterRelation {
+  type: RelationsForConvenienceMethods;
+  text: string;
+}
+export interface ExpectExistenceInputParameter {
```
adi-wan-askui, line 35:
```suggestion
export interface ElementExistsQuery {
```

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -721,136 +677,51 @@ export class UiControlClient extends ApiCommands {
...
    * @property {RelationsForConvenienceMethods} params.relation.type - The type of relation.
    * @property {string} params.relation.text - The label or text associated with the
    *                                           related element or state.
-   * @returns {ExpectExistenceReturnValue} - Property boolean 'everythingExists' tells
-   *                                         if every element exists.
+   * @returns {ExpectExistenceReturnValue.everythingExists} - If every element exists.
+   * @returns {ExpectExistenceReturnValue.elements} - ExpectExistenceElement[].
    */
   async expectExistence(
     params: ExpectExistenceInputParameter[],
```
adi-wan-askui, line 706: queries: ElementExistsQuery[]

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -19,21 +23,27 @@ import { UiControlClientDependencyBuilder } from './ui-control-client-dependency
...
+export interface ExpectExistenceInputParameter {
+  type: keyof Pick<FluentFilters, ExpectExistenceInputParameterType>;
+  text?: ExpectExistenceInputParameterText;
+  relation?: ExpectExistenceInputParameterRelation;
+}
+export interface ExpectExistenceElement extends ExpectExistenceInputParameter {
+  exists: boolean;
+}
 export type ExpectExistenceReturnValue = {
   everythingExists: boolean;
```
adi-wan-askui, line 44: allExist: boolean;

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -19,21 +23,27 @@ import { UiControlClientDependencyBuilder } from './ui-control-client-dependency
...
+}
+export interface ExpectExistenceInputParameter {
+  type: keyof Pick<FluentFilters, ExpectExistenceInputParameterType>;
+  text?: ExpectExistenceInputParameterText;
+  relation?: ExpectExistenceInputParameterRelation;
+}
+export interface ExpectExistenceElement extends ExpectExistenceInputParameter {
+  exists: boolean;
+}
 export type ExpectExistenceReturnValue = {
```
adi-wan-askui, line 43:
```suggestion
export type AllExistsResponse = {
```

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -19,21 +23,27 @@ import { UiControlClientDependencyBuilder } from './ui-control-client-dependency
...
+export interface ExpectExistenceInputParameterRelation {
+  type: RelationsForConvenienceMethods;
+  text: string;
+}
+export interface ExpectExistenceInputParameter {
+  type: keyof Pick<FluentFilters, ExpectExistenceInputParameterType>;
+  text?: ExpectExistenceInputParameterText;
+  relation?: ExpectExistenceInputParameterRelation;
+}
+export interface ExpectExistenceElement extends ExpectExistenceInputParameter {
```
adi-wan-askui, line 40:
```suggestion
export interface ElementExistsResponse extends ElementExistsQuery {
```

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -19,21 +23,27 @@ import { UiControlClientDependencyBuilder } from './ui-control-client-dependency
...
-  label: string;
-  matching?: 'similar' | 'exact' | 'regex';
-  relation?: {
-    type: RelationsForConvenienceMethods;
-    text?: string;
-  };
-  exists?: boolean;
-};
+export type TextMatchingOption = 'similar' | 'exact' | 'regex';
+export type ExpectExistenceInputParameterType = 'otherElement' | 'switch' | 'element' | 'container' | 'checkbox' | 'element' | 'button' | 'table' | 'text' | 'icon' | 'image' | 'textfield';
```
adi-wan-askui, line 26:
```suggestion
export type ElementExistsQueryElementType = 'otherElement' | 'switch' | 'element' | 'container' | 'checkbox' | 'element' | 'button' | 'table' | 'text' | 'icon' | 'image' | 'textfield';
```

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -19,21 +23,27 @@ import { UiControlClientDependencyBuilder } from './ui-control-client-dependency
...
+export interface ExpectExistenceInputParameterText {
```
adi-wan-askui, line 27:
```suggestion
export interface ElementExistsQueryText {
```

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -19,21 +23,27 @@ import { UiControlClientDependencyBuilder } from './ui-control-client-dependency
...
+  value: string;
+  matching?: TextMatchingOption;
+}
+export interface ExpectExistenceInputParameterRelation {
```
adi-wan-askui, line 31:
```suggestion
export interface ElementExistsQueryRelation {
```

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -370,31 +380,21 @@ export class UiControlClient extends ApiCommands {
...
-        commando = command.below().text(text);
-        break;
+        return command.below().text(text);
       case 'contains':
-        commando = command.contains().text(text);
-        break;
+        return command.contains().text(text);
       default:
-        throw new ValidationError('No valid Relation.Type was passed.');
+        return command.nearestTo().text(text);
```
adi-wan-askui, line 395: Why not
```typescript
      case 'nearestTo':
        return command.nearestTo().text(text);
```
to be explicit. 

Reasoning: If the user passes a value that is not allowed (not using TS, at least not properly), I would rather tell the user that there is something wrong with his input by failing instead of creating the assumption that this actually works. Also validation of user input should be done as early as possible. If possible, this is immediately in the first lines of the method called by the user so that in the call stack starting from there you can rely fully on typescript and the type system. There are cases where this is not possible or at least not convenient, especially for more semantical errors, e.g., need to call the database to check if an id is actually unique etc., in which cases you fail later but this is not one of such.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -603,84 +603,37 @@ export class UiControlClient extends ApiCommands {
...
-        commando = command.withExactText(text);
-        break;
+        return command.withExactText(text.value);
       case 'regex':
-        commando = command.withTextRegex(text);
-        break;
+        return command.withTextRegex(text.value);
       default:
-        throw new ValidationError('"matching" property must be "similar", "exact" or "regex"');
+        return command.withText(text.value);
```
adi-wan-askui: Why not
```typescript
case 'similar':
        return command.withText(text);
```
instead of the default.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -721,136 +677,51 @@ export class UiControlClient extends ApiCommands {
...
-   *                                   In case of 'text': The text to be matched for
-   * @property {string} params.matching - Whether the text is matched using similarity,
-   *                                      exact match or a regular expression.
+   * @param {ExpectExistenceInputParameter[]} params - Objects containing the required property
+   *                                                   'type' and the optional properties
+   *                                                   'text' and 'relation'.
+   * @property {string} params.type - The type of the element: 'otherElement' | 'switch' |
+   *                                  'element' | 'container' | 'checkbox' | 'element' |
+   *                                  'button' | 'table' | 'text' | 'icon' | 'image' | 'textfield'
+   * @property {Object} params.text - Object containing value and matching strategy.
```
adi-wan-askui, line 693: Why not be explicit about this and the other (optional) params / params with default values (see https://jsdoc.app/tags-param#optional-parameters-and-default-values)? 🙂

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -721,136 +677,51 @@ export class UiControlClient extends ApiCommands {
...
+   *                                           exact match or a regular expression.
    * @property {Object} params.relation - Object describing the relationship between the
    *                                      clicked text and another element.
    * @property {RelationsForConvenienceMethods} params.relation.type - The type of relation.
    * @property {string} params.relation.text - The label or text associated with the
    *                                           related element or state.
-   * @returns {ExpectExistenceReturnValue} - Property boolean 'everythingExists' tells
-   *                                         if every element exists.
+   * @returns {ExpectExistenceReturnValue.everythingExists} - If every element exists.
+   * @returns {ExpectExistenceReturnValue.elements} - ExpectExistenceElement[].
```
adi-wan-askui: When updating the names, we must not forget to update this.

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -721,136 +677,51 @@ export class UiControlClient extends ApiCommands {
...
-      resolve(returnValue);
-    });
+      return [
+        ...acc,
+        {
+          ...param,
+          exists: (await finalCommand.exec()).length > 0,
+        },
+      ];
+    }, Promise.resolve([] as ExpectExistenceElement[]));
```
adi-wan-askui, line 728: You don't need a type cast. this is a generic (⟶ generics) method (have a look at the signature):
```suggestion
    }, Promise.resolve<ExpectExistenceElement[]>([]));
```

adi-wan-askui (changes requested):
Well done switching to reduce :)! 

To your question: You can replace the typecasts by signature overloads. I provided an example to you inside a commit to this branch. You can replace the other typecast the same way.

Left some last ideas for improving the naming, usage of switch statements and documenting parameters.

packages/askui-nodejs/src/execution/ui-control-client.ts
JohannesDienst-askui: I agree, because we require the user to set a correct relation type 😃 .

packages/askui-nodejs/src/execution/ui-control-client.ts
adi-wan-askui: Actually, these are my reasons for changing it 😆 Parameter is optional and defaults to `similar`. Code currently implies that another value could be possible here as we have a general default clause which is not the case. If someone dropped out typechecking by TS and actually provided a different value, we would simply go with the default here instead of failing and educating the user with the error message. Let's say a user provided `'regexp'` instead of the valid `'regex'`, the user would expect that the regular expression is used instead we use similarity as we don't tell the user nothing about the mistake in using the api.

adi-wan-askui (approved):
Well done :)

PR #761, opened 2024-07-22
Add optional param to exec on shell command

mlikasam-askui:
- Allow parameters to be passed to actions.
- Introduce an optional parameter for the execOnShell command in the shell to specify the expected timeout.

**Jira:** CO-755

### DIFF
--- a/packages/askui-nodejs/src/execution/dsl.ts
+++ b/packages/askui-nodejs/src/execution/dsl.ts
@@ -2418,7 +2418,7 @@ export abstract class FluentCommand extends FluentBase {
   }
 
   /**
-   * Executes a shell command on the device your UiController is connected to.
+   * Executes a shell command on the device your AskUI Controller is connected to.
    *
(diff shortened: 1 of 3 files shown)


packages/askui-nodejs/src/execution/dsl.ts
```diff
@@ -3186,17 +3186,25 @@ export abstract class FluentCommand extends FluentBase {
...
+   * @param {string} shellCommand - A shell command which is executed.
+   * @param {number} [timeoutInMilliseconds=1] - A timeout in milliseconds.
    *
    * @return {Exec}
    */
-  execOnShell(shell_command: string): Exec {
+  execOnShell(
+    shellCommand: string,
+    timeoutInMilliseconds
+    = 1,
```
dom-askui, line 3197: Is this formatting wanted like that?

dom-askui (approved):

PR #762, opened 2024-07-26
feat: add .gitignore; Error log show line of error and code now (#DVRL-81)

JohannesDienst-askui:


packages/askui-nodejs/src/execution/dsl.ts
```diff
@@ -825,7 +411,7 @@ export class FluentFilters extends FluentBase {
    *     .customElement({
    *         customImage: './logo.png', // required
    *         name: 'myLogo', // optional
-   *         threshold: 0.5, // optional, defaults to 0.5
+   *         threshold: 0.9, // optional, defaults to 0.9
```
programminx-askui, line 414: I think you've reverted the default parameter from @adi-wan-askui .

packages/askui-nodejs/src/execution/dsl.ts
```diff
@@ -1935,7 +1552,7 @@ export class FluentFiltersCondition extends FluentBase {
...
```
programminx-askui, line 1555: Same here.

programminx-askui (changes requested):
@mlikasam-askui  has another [PR](https://github.com/askui/askui/pull/761) open, which change the DSL. I we should be carefull regarding merging.

packages/askui-nodejs/example_projects_templates/templates/askui-helper-windows.nj
```diff
@@ -17,12 +17,16 @@ beforeAll(async () => {
 });
 
 beforeEach(async () => {
+  /* Uncomment to enable video recording
   await aui.startVideoRecording();
+  */
```
mlikasam-askui, line 22: What do you think about adding a param to the askui init called `--with-video-recording` and set it to false as a default a value?

programminx-askui: I'm removing video recording for each customer. And nobody has complaint about it. 

The idea was to uncomment it, so the user has the possibility to enable it later.

programminx-askui: And for replacing the Lagacy Controller Video recording needs to be removed.

packages/askui-nodejs/src/execution/dsl.ts
JohannesDienst-askui: Yes, indeed.

packages/askui-nodejs/example_projects_templates/templates/askui-helper-windows.nj
JohannesDienst-askui: Do you think we should merge this or should I add a flag? For me both are fine 😉 .

programminx-askui: Let's merge it.

programminx-askui (approved):

PR #764, opened 2024-08-02
feat: adapt init command to the new installer: No credentials and AskUI Controller setup (#APD-86)

JohannesDienst-askui:
This removes the following from the Interactive CLI and Example project template:

* AskUI Controller setup in `askui-helper.ts`
* Credentials questions

@mlikasam-askui Can you please double-check if this will work with the installer?


packages/askui-nodejs/example_projects_templates/templates/askui-helper.nj
```diff
@@ -43,8 +30,6 @@ afterEach(async () => {
 
 afterAll(async () => {
   aui.disconnect();
-
-  await uiController.stop();
 });
 
-export { aui };
+export { aui };
```
mlikasam-askui, line 35:
```suggestion
export { aui };

```

mlikasam-askui (approved):

PR #793, opened 2024-10-17
fix: escape text strings in askui code in annotation

adi-wan-askui:
- quotation marks, new lines etc.


mlikasam-askui (approved):
nice catch !

PR #801, opened 2024-10-28
feat(ML-595): add custom modelcomposition in exec

danyalxahid-askui:
# Context

Provide model composition for specific actions in exec() command. Model composition in Exec command will take precedence on the modelcomposition set inside `askui-helper.ts`.


programminx-askui (approved):
Seems fine from my side. Did you tested it? And did you also changed this in askui-inference?

danyalxahid-askui:
> Seems fine from my side. Did you tested it? And did you also changed this in askui-inference?

Yes tested it out. The inference is also in [PR](https://github.com/askui/askui-inference/pull/240).

PR #849, opened 2025-01-30
refactor: remove table and matching commands (#CL-1174)

danyalxahid-askui:


packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -483,7 +483,8 @@ export class UiControlClient extends ApiCommands {
    * @param {string} description
    */
   async clickIcon(description: string) {
-    await this.click().icon().matching(description).exec();
+    await this.click().icon().contains().text(description)
+      .exec();
   }
```
adi-wan-askui, line 488: Shouldn't we remove this? I don't think this does not work anymore as expected with icon detection and matching deprecated.

adi-wan-askui (changes requested):
Nice clean up :)

packages/askui-nodejs/src/execution/dsl.ts
```diff
@@ -273,75 +273,6 @@ export class FluentFilters extends FluentBase {
...
-
-    return new FluentFiltersOrRelations(this);
-  }
-
-  /**
-   * Filters for a UI element 'table'.
-   *
-   * @return {FluentFiltersOrRelations}
-   */
-  table(): FluentFiltersOrRelations {
```
adi-wan-askui, line 337: I think we did not deprecate this including the other `table` methods below

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -27,7 +27,7 @@ import { ModelCompositionBranch } from './model-composition-branch';
 
 export type RelationsForConvenienceMethods = 'nearestTo' | 'leftOf' | 'above' | 'rightOf' | 'below' | 'contains';
 export type TextMatchingOption = 'similar' | 'exact' | 'regex';
-export type ElementExistsQueryType = 'otherElement' | 'switch' | 'element' | 'container' | 'checkbox' | 'element' | 'button' | 'table' | 'text' | 'icon' | 'image' | 'textfield';
+export type ElementExistsQueryType = 'otherElement' | 'switch' | 'element' | 'container' | 'checkbox' | 'element' | 'button' | 'text' | 'icon' | 'image' | 'textfield';
 export interface ElementExistsQueryText {
   value: string;
   matching?: TextMatchingOption;
```
adi-wan-askui, line 33: Do we still support this?

packages/askui-nodejs/src/execution/ui-control-client.ts
```diff
@@ -27,7 +27,7 @@ import { ModelCompositionBranch } from './model-composition-branch';
 
 export type RelationsForConvenienceMethods = 'nearestTo' | 'leftOf' | 'above' | 'rightOf' | 'below' | 'contains';
 export type TextMatchingOption = 'similar' | 'exact' | 'regex';
-export type ElementExistsQueryType = 'otherElement' | 'switch' | 'element' | 'container' | 'checkbox' | 'element' | 'button' | 'table' | 'text' | 'icon' | 'image' | 'textfield';
```
adi-wan-askui, line 30: I don't think `table` is gone or is it?

adi-wan-askui (changes requested):

packages/askui-nodejs/src/execution/ui-control-client.ts
danyalxahid-askui: Yes. By default we use similar. But if we provide exact or regex we use that matching option.

packages/askui-nodejs/src/execution/ui-control-client.ts
danyalxahid-askui: removed.

packages/askui-nodejs/src/execution/dsl.ts
danyalxahid-askui: table is still supported. Row, Column, cells are not supported now.

adi-wan-askui (approved):

PR #854, opened 2025-04-16
Disable retry strategy for waitUntil command

mlikasam-askui:


programminx-askui (approved):
We will move the Support RetryStrategy in exec functions

```
exec(retryStrategy?:  RetryStrategy)
```

PR #877, opened 2026-04-08
fix(http-client): enable got retry for transient server errors on POS…

programminx-askui:
…T requests (#SOLENG-333)

Remove `throwHttpErrors: false` from the POST method so that got's built-in retry mechanism (5 retries, exponential backoff) can trigger for transient server errors (500, 502, 503, 504). Previously, got silently returned error responses without retrying because it did not see them as errors. After retries are exhausted, HTTPError is caught and converted to our custom error hierarchy for consistent handling.

This fixes intermittent CUDNN_STATUS_INTERNAL_ERROR (500) responses from the Azure GPU backend that cannot be fixed server-side.

Refs: SOLENG-333

### DIFF
--- a/packages/askui-nodejs/src/utils/http/http-client-got.ts
+++ b/packages/askui-nodejs/src/utils/http/http-client-got.ts
@@ -1,6 +1,7 @@
 import got, {
   ExtendOptions,
   Got,
+  HTTPError,
   OptionsOfJSONResponseBody,
   RequestError,
   TimeoutError,
@@ -57,6 +58,16 @@ export class HttpClientGot {
     this.askuiGot = got.extend(gotExtendOptions);
   }
 
+  /**
+   * Configures got with retry behavior for transient server errors.
+   *
+   * Got retries requests that fail with retryable status codes
+   * (500, 502, 503, 504, etc.) up to `limit` times using exponential backoff.
+   *
+   * POST requests are only retried if their URL is registered in `urlsToRetry`
+   * (see `shouldRetryOnError`), to avoid retrying non-idempotent calls
+   * to unknown endpoints.
+   */
   private buildGotExtendOptions(
     proxyAgents?: { http: http.Agent; https: https.Agent } | undefined,
   ): ExtendOptions {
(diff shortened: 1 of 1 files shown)


packages/askui-nodejs/src/utils/http/http-client-got.ts
```diff
@@ -57,6 +58,16 @@ export class HttpClientGot {
     this.askuiGot = got.extend(gotExtendOptions);
   }
 
+  /**
+   * Configures got with retry behavior for transient server errors.
+   *
+   * Got retries requests that fail with retryable status codes
+   * (500, 502, 503, 504, etc.) up to `limit` times using exponential backoff.
+   *
+   * POST requests are only retried if their URL is registered in `urlsToRetry`
```
mlikasam-askui, line 68: urlsToRetry is empty

packages/askui-nodejs/src/utils/http/http-client-got.ts
```diff
@@ -187,22 +203,39 @@ export class HttpClientGot {
...
-      throw httpClientErrorHandler(statusCode, JSON.stringify(body));
+    try {
+      const { body, statusCode, headers } = await this.askuiGot.post<T>(
+        url,
+        options,
+      );
+      if (headers['deprecation'] !== undefined) {
+        logger.warn(headers['deprecation']);
+      }
+      if (statusCode !== 200) {
```
mlikasam-askui: Why did you set throwHttpErrors to true? Isn’t throwing on all non-200 responses sufficient?

packages/askui-nodejs/src/utils/http/http-client-got.ts
programminx-askui: Claude Code and I cross checked it: 

urlsToRetry is not empty at the time requests are made. It's initialized as an empty array in http-client-got.ts:48, but gets populated by InferenceClient in its constructor (inference-client.ts:48):


this.httpClient.urlsToRetry = Object.values(this.urls);
This sets it to all four inference endpoint URLs (inference, is-image-required, vqa/inference, act/inference) before any HTTP requests are sent.

The dependency builder constructs HttpClientGot first, then passes it into InferenceClient, which fills urlsToRetry in its constructor — so by the time post() is called, the URLs are registered and got's shouldRetryOnError will match them.

mlikasam-askui (approved):

